PR #21, opened 2022-07-12
runtime: prepare for beta release

### LINKED ISSUE #19: runtime: Update parameters before release
* Use a more recent Testnet trust root with the correct runtime ID.
* Set `debug = false` in `Cargo.toml` under `package.metadata.fortanix-sgx`.
* Configure proper runtime IDs in `Cargo.toml` under `package.metadata.orc.*`.
* Check if configured `max_tx_size` is ok.
* Change version to `0.1.0-testnet`.

nhynes:
resolves #19


runtime/Cargo.toml
```diff
@@ -1,26 +1,26 @@
...
 license = "Apache-2.0"
-resolver = "2"
 
 [package.metadata.orc.release]
-runtime-id = "000000000000000000000000000000000000000000000000e2eaa99fc008f87f"
+runtime-id = "000000000000000000000000000000000000000000000000a6d1e3ebf60dff6c"
 
 [package.metadata.orc.test]
-runtime-id = "00000000000000000000000000000000000000000000000072c8215e60d5bca7"
+runtime-id = "0000000000000000000000000000000000000000000000000000000000000000"
```
kostko, line 12: This is the ID used on testnet, so it should be the one above. The mainnet one is not yet defined, so just switch the two.

runtime/src/lib.rs
```diff
@@ -94,7 +95,11 @@ impl sdk::Runtime for Runtime {
     );
 
     fn trusted_policy_signers() -> Option<TrustedPolicySigners> {
-        Some(cipher_keymanager::trusted_policy_signers())
+        let tps = cipher_keymanager::trusted_policy_signers();
+        Some(TrustedPolicySigners {
+            signers: tps.signers.into_iter().map(|s| PublicKey(s.0)).collect(),
```
kostko, line 100: I assume this is due to the version mismatch? Just add a comment saying so.

runtime/src/lib.rs
```diff
@@ -118,7 +123,7 @@ impl sdk::Runtime for Runtime {
                 parameters: modules::core::Parameters {
                     min_gas_price: { BTreeMap::from([(Denomination::NATIVE, 100_000_000_000)]) },
                     max_batch_gas: if is_testnet() { 30_000_000 } else { 10_000_000 },
-                    max_tx_size: 32 * 1024,
+                    max_tx_size: u32::max_value(),
```
kostko, line 126: This should be reduced to something more reasonable?

kostko: AFAIK there is a limit for p2p propagation? If you rely on gas alone then the per-byte cost needs to be increased. I would just set this to something like 300 KB or so?

nhynes: I haven't encountered the p2p propagation issue in the same way that I haven't run out of gas due to tx size alone. 300 KiB is probably fine. I'd prefer to have cheap txs in the common case anyway.

kostko (approved):

PR #27, opened 2022-08-04
clients/js: add compat lib

nhynes:
This ends up as

```typescript
import { wrap } from '@oasisprotocol/sapphire-paratime';

// ethers
import { Wallet, Web3Provider } from 'ethers';
const signer = wrap(new Web3Provider(window.ethereum).getSigner());
const signer = wrap(new Wallet('private key'));

// web3.js (used by truffle)
import Web3 from 'web3';
web3.setProvider(wrap(web3.currentProvider));

// raw
const eth = wrap(window.ethereum);
```

The unit tests work and follow the spec.. E2E tests are still TODO, but those will arrive eventually.

Screen Shot 2022-08-19 at 01 13 11


clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+  chainId: number,
+  opts?: ProviderOptions,
+): Promise<KeyManagerSignedPublicKey> {
+  const cached = PK_CACHE.get(chainId);
+  if (cached) return cached;
+
+  const rtId = opts?.runtimeId ?? inferRuntimeId(chainId);
+  const rpc = opts?.rpc ?? inferRpcUrl(chainId);
+  const { public_key: publicKey } = await new CoreWrapper(arrayify(rtId))
+    .queryCallDataPublicKey()
```
kostko, line 269: Should this somehow be exposed via a Web3 REST API extension to avoid the need for two separate connections?

nhynes: it absolutely should, but we need to cut an alpha of this lib asap

clients/js/package.json
```diff
@@ -11,34 +11,44 @@
   ],
   "type": "module",
   "main": "lib/index.js",
+  "module": "lib/index.js",
   "types": "lib/index.d.ts",
+  "exports": {
+    ".": "lib/index.js",
```
xmzheng, line 17: I got this error when I tried to import this module:
```
Error [ERR_INVALID_PACKAGE_TARGET]: Invalid "exports" main target "lib/index.js" defined in the package config /Users/jameszheng/Documents/Oasis/ts-deployment/node_modules/@oasisprotocol/sapphire-paratime/package.json imported from /Users/jameszheng/Documents/Oasis/ts-deployment/build/index.js; targets must start with "./"
```

Do we need
```suggestion
    ".": "./lib/index.js",
```

clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+  TransactionRequest,
+  Web3Provider,
+  JsonRpcSigner,
+} from '@ethersproject/providers';
+import * as cbor from 'cborg';
+
+import callformat, {
+  MetaEncryptedX25519DeoxysII,
+} from '@oasisprotocol/client-rt/dist/callformat.js';
+import { NodeInternal } from '@oasisprotocol/client/dist/client';
```
xmzheng, line 23:
```suggestion
import { NodeInternal } from '@oasisprotocol/client/dist/client.js';
```

clients/js/package.json
```diff
@@ -14,7 +14,7 @@
   "module": "lib/index.js",
   "types": "lib/index.d.ts",
   "exports": {
-    ".": "lib/index.js",
+    ".": "./lib/index.js",
     "./compat": "lib/compat.js",
     "./signed_calls": "lib/signed_calls.js"
```
xmzheng, line 19: these two lines too?

clients/js/src/signed_calls.ts
```diff
@@ -15,6 +15,8 @@ import type {
   SnakeCasedProperties,
 } from 'type-fest';
 
+import { MAINNET_CHAIN_ID, TESTNET_CHAIN_ID } from '.';
```
xmzheng, line 18:
```suggestion
import { MAINNET_CHAIN_ID, TESTNET_CHAIN_ID } from './index.js';
```

clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+import { EthCall, prepareSignedCall } from './signed_calls.js';
+import {
+  MAINNET_CHAIN_ID,
+  MAINNET_RT_ID,
+  TESTNET_CHAIN_ID,
+  TESTNET_RT_ID,
+} from './index.js';
+
+const TESTNET_RPC = 'https://grpc.oasis.dev';
+const MAINNET_RPC = 'https://testnet.grpc.oasis.dev';
```
lukaw3d, line 37: swapped names

clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+              provider.call.bind(provider) as EthersCall<string>,
+            );
+          }
+          if (prop === 'estimateGas') {
+            return hookEthersCall(
+              signer,
+              provider.estimateGas.bind(provider) as EthersCall<BigNumber>,
+            );
+          }
+          return Reflect.get(target, prop).bind(target);
```
xmzheng, line 137:
```suggestion
          let value = Reflect.get(target, prop);
          return typeof value == 'function' ? value.bind(target) : value;
```

clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+        },
+      }),
+    ) as T;
+  }
+
+  return new Proxy(upstream, {
+    get(target, prop) {
+      if (prop !== 'request') return Reflect.get(target, prop).bind(target);
```
xmzheng, line 145: same here.

clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+async function fetchRTPubKey(
+  chainId: number,
+  opts?: ProviderOptions,
+): Promise<KeyManagerSignedPublicKey> {
+  const cached = PK_CACHE.get(chainId);
+  if (cached) return cached;
+
+  const rtId = opts?.runtimeId ?? inferRuntimeId(chainId);
+  const rpc = opts?.rpc ?? inferRpcUrl(chainId);
+  const { public_key: publicKey } = await new CoreWrapper(arrayify(rtId))
```
xmzheng, line 268:
```suggestion
  const { public_key: publicKey } = await new CoreWrapper(arrayify(rtId, {allowMissingPrefix: true}))
```

clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+        get(target, prop) {
+          if (prop === 'sendTransaction') {
+            return (async (signedTx) => {
+              const [ptTx, chainId] = await Promise.all([
+                signedTx,
+                signer.getChainId(),
+              ]);
+              const [encTx] = await encrypt(ptTx, chainId, opts);
+              return provider.sendTransaction(encTx);
+            }) as EthersProvider['sendTransaction'];
```
xmzheng, line 123: The wrapper logic doesn't sound correct to me. It seems that we should encrypt the transaction data first and then sign it and finally send it out. Shouldn't we Proxy `EthersSigner['sendTransation']` instead of `EthersProvider['sendTransaction']` here?

nhynes: > Shouldn't we proxy [...]

Encryption happens at the lowest layer (the provider) because we intend to make it possible for proxies in the middle of the exchange to encrypt the transaction.

nhynes: This isn't true for signed queries, though, which I'd agree is odd, but it's weird to have something like:

```js
const origCall  = {
    gas: 42,
    gasLimit: 1000,
    from: '0xabc...',
    to: '0xdef..',
    data: '0xffff',
};
const encCall = {
  data: enc(origCall.data),
  ...origCall,
};
const signedCall = {
  data: encode({
    data: encCall.data,
    signature: sign(encode(encCall))
  }),
  ...encCall
};
```

versus now where it's
```
const signedCall = {
  call: encCall,
  signature: sign(encode(encCall))
}
```
which is more normal in terms of signatures, but doesn't permit proxy encryption

xmzheng: > Encryption happens at the lowest layer (the provider) because we intend to make it possible for proxies in the middle of the exchange to encrypt the transaction.

I just didn't understand how can we encrypt a signed transaction, which would invalidate the signatures. 

https://github.com/ethers-io/ethers.js/blob/01aea705ce60b1c42d2f465b162cb339a0e94392/packages/abstract-signer/src.ts/index.ts#L125

clients/js/src/compat.ts
```diff
@@ -0,0 +1,289 @@
...
+async function encrypt(
+  data: BytesLike,
+  chainId: number,
+  opts: ProviderOptions | undefined,
+): Promise<[string, MetaEncryptedX25519DeoxysII?]> {
+  if (data.length === 0) return [hexlify(data)];
+
+  const [packed, meta] = await callformat.encodeCall(
+    {
+      method: '',
```
xmzheng, line 212:
```suggestion
      format: CALLFORMAT_ENCRYPTED_X25519DEOXYSII,
      method: '',
```

clients/js/src/compat.ts
xmzheng: > 
```
const signedCall = {
  data: encode({
    data: encCall.data,
    signature: sign(encode(encCall))
  }),
  ...encCall
};
```

This format sounds more natural to me. The second format is also fine since we have already handled that in our evm module. The format here doesn't really matter since the `data` field in a call request doesn't involve any sign process at the wallet level.

xmzheng: https://testnet.explorer.sapphire.oasis.dev/tx/0x7ab6527a057c486fc327ef6ca65a34b5d1b40f360452b88146736aaad4a6923c/token-transfers

I changed the proxy to Signer['sendTransaction'] and use 'CALLFORMAT_PLAIN', it deployed successfully.

```
  if (EthersSigner.isSigner(upstream) && '_signTypedData' in upstream) {
    const signer: EthersSigner & TypedDataSigner = upstream;
    const provider = signer.provider;
    if (provider === undefined) {
      throw new Error(
        'ethers Signer must be connected to a Provider before wrapping',
      );
    }
    return new Proxy(signer as EthersSigner, {
      get(target, prop) {
	 if (prop === 'sendTransaction') {
	   return (async (originTx) => {
	     const chainId = await signer.getChainId();
             const [encTx] = await encrypt(originTx.data as string, chainId, opts);
	     originTx.data = encTx;
             return signer.sendTransaction(originTx);	     
	   }) as EthersSigner['sendTransaction'];
	 }
	let value = Reflect.get(target, prop);
	return typeof value == 'function' ? value.bind(target) : value;
      }
    }) as T;
  }
```

clients/js/src/compat.ts
nhynes: this one doesn't need to be set. the inner one is plain, which is the default

clients/js/src/compat.ts
nhynes: > I just didn't understand how can we encrypt a signed transaction, which would invalidate the signatures.

Encryption doesn't invalidate signatures. You just need to unwrap the encryption before verifying, if needed.

pro-wh (approved):
👍 

other reviewers please review if you have experience with the ethers library

the proxy stuff is a little spooky, but it should work as long as none of the following happen:
1. any code does wrappedProvider.somethingWriteable = value and expects to be able to read it back
2. any code does Object.create(wrappedProvider)
3. the upstream provider has any methods that call into something that ought to be hooked, e.g. logAndSend(args, cb) { log(args); this.send(args, cb); }

clients/js/src/cipher.ts
```diff
@@ -0,0 +1,311 @@
...
+}
+
+export function lazy(generator: () => Promisable<Cipher>): Cipher {
+  return new Proxy(
+    {},
+    {
+      get(target: { inner?: Cipher }, prop) {
+        // Props (Promiseable)
+        if (prop === 'kind' || prop === 'publicKey') {
+          if (target.inner) return Reflect.get(target.inner, prop);
```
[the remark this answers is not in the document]
nhynes: yeah, it'll complain about not having an index signature otherwise

clients/js/src/cipher.ts
```diff
@@ -0,0 +1,311 @@
...
+          return Promise.resolve(generator()).then((inner: Cipher) => {
+            target.inner = inner;
```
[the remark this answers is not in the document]
nhynes: It'd be nice if `Promise`s were lazy themselves.

nhynes:
Thanks for the review Warren.

To address your caveats, the props on the providers are (intended to be) read-only. I've never seen anyone set stuff on a provider. All of the ethers providers almost all people use are in `@ethersproject/providers`, which don't have any of the behaviors you mention. Web3.js is even simpler. If anyone uses a truly custom provider, I'd love a bug report.

PR #29, opened 2022-08-05
runtime: debug builds

nhynes:
This PR has the release CI workflow generate debug ORCs that can be run with any keymanager.

related: https://github.com/oasisprotocol/oasis-core/issues/4876

The CI step is tested [here](https://github.com/nhynes/sapphire-paratime/runs/7692156509?check_suite_focus=true).

There's an alternative approach where debug ORCs are also part of artifacts and don't need signing, but those only last 90 days and are not suitable for embedding in CI steps.


runtime/src/lib.rs
```diff
@@ -95,6 +95,9 @@ impl sdk::Runtime for Runtime {
     );
 
     fn trusted_policy_signers() -> Option<TrustedPolicySigners> {
+        if option_env!("OASIS_UNSAFE_SKIP_KM_POLICY") == Some("1") {
```
kostko, line 98: The other places just check whether the env variable is present, but this works too.

.github/workflows/release.yaml
```diff
@@ -23,6 +17,39 @@ jobs:
...
+        uses: oasisprotocol/oasis-sdk/.github/actions/hash-rust@main
+        with:
+          image: oasisprotocol/runtime-builder:main
+          pkg-dirs: runtime
+          binaries: sapphire-paratime
+          clean: no
+          setup: |
+            export OASIS_UNSAFE_SKIP_AVR_VERIFY=1 OASIS_UNSAFE_SKIP_KM_POLICY=1 OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES=1
+
+      - name: Build tagged release, sgxs, debug
```
kostko, line 31: Any reason for also including a debug SGXS build? I assume nobody will sign this? You can skip the SGXS variant if this is only used for local testing (just omit the `--sgx-executable` when building an ORC.

.github/workflows/release.yaml
```diff
@@ -57,5 +84,5 @@ jobs:
         with:
           # Create a draft since the release requires an offline signing process.
           draft: true
-          artifacts: sapphire-paratime.orc
+          artifacts: sapphire-paratime.orc,debug-sapphire-paratime.orc
```
kostko, line 87:
```suggestion
          artifacts: sapphire-paratime.orc,unsafe-debug-sapphire-paratime.orc
```

.github/workflows/release.yaml
```diff
@@ -23,6 +17,39 @@ jobs:
...
+          clean: no
+          sgx: yes
+          setup: |
+            export OASIS_UNSAFE_SKIP_AVR_VERIFY=1 OASIS_UNSAFE_SKIP_KM_POLICY=1 OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES=1
+
+      - name: Create the debug Oasis Runtime Containers
+        run: |
+          go install github.com/oasisprotocol/oasis-sdk/tools/orc@latest
+          pushd runtime
+          orc init ${RUNTIME_EXECUTABLE} --sgx-executable ${RUNTIME_SGXS} --output ../debug-sapphire-paratime.orc
```
kostko, line 47:
```suggestion
          orc init ${RUNTIME_EXECUTABLE} --sgx-executable ${RUNTIME_SGXS} --output ../unsafe-debug-sapphire-paratime.orc
```

.github/workflows/release.yaml
nhynes: it'd be cool to someday have a key in CI that auto-signs this debug build in a way that's insecure but convenient. Then it'd be possible to test in real SGX

kostko: You don't need to sign debug builds, Oasis Core will do it for you using a known (and blacklisted in production) signer if configured in unsafe mode. If you think debug SGX builds are also useful, use it as-is.

kostko: Oh, also if you want debug SGX builds, the other problem is that the `consensus_trust_root` binds the runtime to the given consensus layer deployment so this will not work in a local environment.

kostko (approved):

nhynes:
auto-merge strikes again? I think I should just stop using it or else wait forever for reviews :P

PR #33, opened 2022-08-19
clients/js: fix memoization

nhynes:
There was a follow-up comment on #27


pro-wh (approved):
in cases when we run two copies of generate(), with this change, we'll keep the first fulfillment.

PR #34, opened 2022-08-23
clients/js: improve web3.js compat

nhynes:


kostko (approved):

clients/js/src/compat.ts
```diff
@@ -123,17 +123,17 @@ export function wrap<U extends UpstreamProvider>(
       cb: (err: unknown, ok?: unknown) => void,
     ) => {
       hookP(args)
-        .then((res) => cb(null, { result: res }))
+        .then((res) => cb(null, { jsonrpc: '2.0', id: 1, result: res }))
```
xmzheng, line 126: `id: res.id`

xmzheng:
```
interface Web3ReqArgs {
  readonly id?: number;
  readonly method: string;
  readonly params?: any[];
}
```

PR #37, opened 2022-08-30
Bump Oasis SDK to rev b8e5f8f8c725eaf28f643f221930b55775b5a765

kostko:


nhynes (approved):

runtime/src/lib.rs
```diff
@@ -95,7 +95,7 @@ impl sdk::Runtime for Runtime {
     );
 
     fn trusted_policy_signers() -> Option<TrustedPolicySigners> {
-        if option_env!("OASIS_UNSAFE_SKIP_KM_POLICY") == Some("1") {
+        if option_env!("OASIS_UNSAFE_SKIP_KM_POLICY").is_some() {
```
nhynes, line 98: strictly speaking, I think it's the other code that's not ideal. `OASIS_UNSAFE_SKIP_KM_POLICY=no` seems like not a great way to turn it on.

kostko: Besides being consistent it also has the side effect of shutting up clippy. Let me add an ignore instead.

PR #52, opened 2022-09-21
Add tutorial for simple e-commerce dApp with confidential payment

xmzheng:


tjanez (commented):
Huh, this is very large (+200k lines added to the repo)...

I think it would be better to create a standalone repo (e.g. `sapphire-store-example`) which people could easily clone to start working on their new dApp and add the tutorial text to our docs repo as an additional guide inside https://docs.oasis.io/dapp/sapphire/.

nhynes:
I'm going to close this issue to clean up the tracker. Please reopen if needed.

matevz:
I went through the PR and most of the diff are compiled contracts ABIs and React leftovers which probably shouldn't be in git anyway? Let's stick to smaller, confidential-specific examples, stress out important details you need to be aware of when building Sapphire dapps and provide boilerplate code, if needed.

PR #54, opened 2022-09-24
TheGraph that is working with Sapphire

kittycatdao:


graph/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
+data/
```
nhynes, line 2: Please add newlines at the ends of files.

graph/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:16.15
+
+WORKDIR /app
+
+COPY package.json yarn.lock ./
+RUN yarn
```
nhynes, line 6:
```suggestion
RUN yarn install --production --frozen-lockfile
```

Strictly speaking, I'd prefer pnpm like in the rest of this repo, but you'd have to get that into the container. If you've got a bit of free time, could you [add pnpm using corepack](https://pnpm.io/installation#using-corepack) and use it instead of yarn?

graph/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+      - '127.0.0.1:8001:8001'
+      - '127.0.0.1:8020:8020'
+      - '127.0.0.1:8030:8030'
+      - '127.0.0.1:8040:8040'
+    extra_hosts:
+      - host.docker.internal:host-gateway
+    environment:
+      postgres_host: postgres
+      postgres_user: graph-node
+      postgres_pass: let-me-in
```
nhynes, line 30: Is this intended to be deployed? If not, could you please rename this directory to `example-graph` or something that evidences that it's not a good idea to deploy this directly? Either that or add a `README.md` about how to make it production-ready.

graph/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "sapphire-proxy",
```
nhynes, line 2:
```suggestion
  "name": "@oasisprotocol/sapphire-graph-proxy",
```

graph/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "version": "1.0.0",
```
nhynes, line 3: Do me a solid and add `"type": "module"` and then use ESM in `index.js`?

graph/package.json
```diff
@@ -0,0 +1,12 @@
...
+  "main": "index.js",
+  "license": "MIT",
+  "dependencies": {
+    "@oasisprotocol/sapphire-paratime": "^1.0.5",
+    "body-parser": "^1.20.0",
```
nhynes, line 8: This comes with Express 4. You can do `express.json()` now.

graph/index.js
```diff
@@ -0,0 +1,33 @@
+const express = require('express');
```
nhynes, line 1: Please use ESM imports like

```js
import express from 'express';
import { ethers } from 'ethers';
import * as sapphire from '@oasisprotocol/sapphire-paratime';
```

graph/index.js
```diff
@@ -0,0 +1,33 @@
...
+const bodyParser = require('body-parser');
+const ethers = require('ethers');
+const sapphire = require('@oasisprotocol/sapphire-paratime');
+
+const app = express();
+app.use(bodyParser.json());
```
nhynes, line 7:
```suggestion
app.use(express.json());
```

graph/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+const provider = new ethers.providers.JsonRpcProvider(process.env.RPC);
+const signer = new ethers.Wallet(Buffer.from(ethers.utils.randomBytes(32)).toString('hex'), provider);
```
nhynes, line 10:
```suggestion
const signer = ethers.Wallet.createRandom();
```

graph/index.js
```diff
@@ -0,0 +1,33 @@
...
+const client = sapphire.wrap(signer);
+
+app.post('/', async (req, res) => {
+    if (req.body.jsonrpc !== '2.0') {
+        res.status(405);
+        res.end();
```
nhynes, line 16:
```suggestion
        res.status(405).end();
```

graph/index.js
```diff
@@ -0,0 +1,33 @@
...
+        return;
+    }
+
+    let payload;
+    if (req.body.method === 'eth_call') {
+        delete req.body.params[0]['gas'];
+        payload = await client.call(req.body.params[0], req.body.params[1]);
+    } else {
+        payload = await provider.send(req.body.method, req.body.params);
+    }
```
nhynes, line 26: this is cosmetic
```suggestion
    let result;
    let { id, method, params } = req.body;
    if (method === 'eth_call') {
        delete params[0]['gas'];
        result = await client.call(...params);
    } else {
        result = await provider.send(method, params);
    }
```

graph/index.js
```diff
@@ -0,0 +1,33 @@
...
+    
+    res.status(200);
+    res.json({ id: req.body.id, jsonrpc: "2.0", result: payload });
+    res.end();
```
nhynes, line 30:
```suggestion
    res.status(200)
        .json({ id, jsonrpc: "2.0", result });
        .end();
```

graph/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "sapphire-proxy",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
```
nhynes, line 5: Could you also add linters (prettier) and formatters (eslint) as in the `clients/js` repo? I know this is a small proxy, but it helps to have them if this is supposed to be long-lived software. You can move the prettierrc from the clients repo to the root.

nhynes (changes requested):
Thanks for the PR! Please let me know if you have any questions about my comments. I'd love to get this merged in.

graph/index.js
kittycatdao: Need to connect signer to JsonRpcProvider then I guess

graph/docker-compose.yml
kittycatdao: I think it might be a good idea to move it out to foreign .env file

graph/Dockerfile
kittycatdao: Will check it out, never used pnpm before tbh

graph/package.json
kittycatdao: Maybe I am missing something but isn't sapphire client lib using CommonJS?

graph/index.js
nhynes: Sorry, I meant `ethers.Wallet.createRandom().connect(...)`. In fact, you don't really need a signer. You can set the `from` address to `address(0)`. I think that'll work in this case, actually.

graph/package.json
kittycatdao: > It [provides both](https://github.com/oasisprotocol/sapphire-paratime/blob/main/clients/js/package.json#L25-L31). Older version of Node don't support ESM, but the latest LTS does, and since ESM is the rapidly advancing standard, we should prefer to use it.

Well, I am not really experienced with ESM outside of TypeScript but I am getting following error when importing oasis client lib:
image

But if I remove import of Oasis lib and keep other imports - it works:
image

So I am a bit confused with this tbh 😅

nhynes: Ugh, node doesn't like the `js` because `type: module` isn't set in the `sapphire-paratime` package. I know that if I do that, hardhat will start complaining. Typescript doesn't want to mess with file extensions, so I guess I have to do it myself? Give me a little bit and I'll get your PR wrapped up :)

kittycatdao: > Ugh, node doesn't like the `js` because `type: module` isn't set in the `sapphire-paratime` package. I know that if I do that, hardhat will start complaining. Typescript doesn't want to mess with file extensions, so I guess I have to do it myself? Give me a little bit and I'll get your PR wrapped up :)

So can I resolve this conversation about ESM?

nhynes: Keep it open for a bit please. If you notice #57, I'll be able to add ESM to your PR and resolve this comment myself.

nhynes:
@kittycatdao can you take a look at the PR and let me know if it still works well for you? I think it's ready to merge

nhynes (approved):

nhynes:
Actually, can you also apply this patch and tell me if it still works? It _should_ since you don't need to construct a signer to wrap.
```patch
diff --git a/integrations/graph/index.js b/integrations/graph/index.js
index 655bcda..0eaa8c3 100644
--- a/integrations/graph/index.js
+++ b/integrations/graph/index.js
@@ -5,9 +5,10 @@ import express from 'express';
 const app = express();
 app.use(express.json());
 
-const provider = new ethers.providers.JsonRpcProvider(process.env.RPC);
-const signer = ethers.Wallet.createRandom().connect(provider);
-const client = sapphire.wrap(signer);
+const provider = sapphire.wrap(
+  new ethers.providers.JsonRpcProvider(process.env.RPC),
+);
+const nobody = Buffer.alloc(20);
 
 app.post('/', async (req, res) => {
   if (req.body.jsonrpc !== '2.0') {
@@ -18,8 +19,8 @@ app.post('/', async (req, res) => {
   let result;
   let { id, method, params } = req.body;
   if (method === 'eth_call') {
-    delete params[0]['gas'];
-    result = await client.call(params[0], params[1]);
+    const { gas, ...txCall } = params[0];
+    result = await wrappedProvider.call({ from: nobody, ...txCall }, params[1]);
   } else {
     result = await provider.send(method, params);
   }
```

kittycatdao:
> @kittycatdao can you take a look at the PR and let me know if it still works well for you? I think it's ready to merge

Works well!

kittycatdao:
Pushed one small fix

nhynes:
Thanks. Did this patch https://github.com/oasisprotocol/sapphire-paratime/pull/54#issuecomment-1261200453 work for you or not? If so, could you apply it and we can call it a PR? :)

kittycatdao:
> Thanks. Did this patch [#54 (comment)](https://github.com/oasisprotocol/sapphire-paratime/pull/54#issuecomment-1261200453) work for you or not? If it did, could you apply it and we can call it a PR? :)

Ah sorry, latest changes don't work, forgot to rebuild docker images last time. I am getting error that oasis lib can't be found as a module and seems it makes sense since there is no "lib" folder:
image

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/alice/Desktop/sapphire-paratime/integrations/graph/node_modules/@oasisprotocol/sapphire-paratime/lib/esm/index.js' imported from /Users/alice/Desktop/sapphire-paratime/integrations/graph/index.js

nhynes:
Good catch! I tested the docker build this time. The package.json had a relative path in it that didn't work in docker (of course). I also applied the patch, so you can just go ahead and test the `eth_call` methods.

kittycatdao:
Pushed fix

Issue #56, opened 2022-09-27
JS client can't be used with CRA


BerkliumBirb:
When trying to import package in Create React App project in development this build error emerges:

```
./node_modules/@oasisprotocol/sapphire-paratime/lib/esm/cipher.js 246:31
Module parse failed: Unexpected token (246:31)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|   }
| 
>   const fetchImpl = globalThis?.fetch ?? opts?.fetch;
|   const res = await (fetchImpl ? fetchRuntimePublicKeyBrowser(gatewayUrl, fetchImpl) : fetchRuntimePublicKeyNode(gatewayUrl));
|   return arrayify(res.result.key);
```

When trying to build production build the following error:
```
Creating an optimized production build...
Failed to compile.

./node_modules/@oasisprotocol/sapphire-paratime/lib/esm/cipher.js
Cannot find module: 'node:https'. Make sure this package is installed.

You can install this package by running: npm install node:https.
```

CRA version: 4.0.3 ([same as in Uniswap](https://github.com/Uniswap/interface/blob/main/package.json#L112))
Node Version: 16.10.0
Sapphire Paratime Version: 1.0.5

BerkliumBirb:
Suggestions to improve this:
1. Downgrade build target so that most build environments doesn't require special configuration for parsing package internals
2. Use something like [cross-fetch](https://www.npmjs.com/package/cross-fetch) to not depend on runtime imports that might puzzle webpack off.

nhynes:
_CRA 4.0.3_? That's ancient 😰 but if the motto is 10 minutes or less, that's just what we've got to do!

BerkliumBirb:
> _CRA 4.0.3_? That's ancient :cold_sweat:

Yeah, but it falls to the same babel-loader with same [babel-preset-react-app/dependencies](https://www.npmjs.com/package/babel-preset-react-app). So updating to v5 will not fix that (but will break `node:https` and `node:crypto` entirely)

PR #58, opened 2022-09-27
Use cross-fetch instead of node:https

BerkliumBirb:
This simplifies bundler setup as it doesn't require to set aliases oasisprotocol/sapphire-paratime#56


nhynes:
I believe that you can replace `node:https` with simply `node` and it'll work. Older verisons of webpack aren't familiar with the new syntax. You could add a loader, but that's too annoying. Can you amend your PR to `https`? I'd rather not pull in a dep if possible, for security reasons.

nhynes:
superseded by #59. thanks for the PR!

PR #60, opened 2022-09-27
Update README.md

Tharkunas:
How to download Rust was missing, I added this

### DIFF
--- a/runtime/README.md
+++ b/runtime/README.md
@@ -50,6 +50,12 @@ The rustup-installed versions of `cargo`, `rustc` and other tools will
 toolchain][rust-toolchain-precedence] when invoked from the Sapphire ParaTime
 git checkout directory.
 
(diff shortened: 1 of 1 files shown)


runtime/README.md
```diff
@@ -50,6 +50,12 @@ The rustup-installed versions of `cargo`, `rustc` and other tools will
 toolchain][rust-toolchain-precedence] when invoked from the Sapphire ParaTime
 git checkout directory.
 
+If rust toolchain is not installed on your device, you can install it like this:
+```
```
nhynes, line 54:
```suggestion
If rust toolchain is not installed on your device, you can install it like this:

```
```

nhynes (approved):
Thanks, this definitely could have been made more clear.

Tharkunas:
No problem mate. We are here to contribute Oasis

PR #62, opened 2022-09-28
runtime: bump version

nhynes:
let's get that 0.1.5-testnet release going!


kostko (commented):
So this will be for the future release right? Because 0.1.4-testnet is the release that is currently being made (we already bumped in the past but did not make a release then) ;)

nhynes:
oh, weird. it makes more sense to have the version be the one you're working on, but it's also hard to know what version you're working on. let's just close this for now :)

PR #68, opened 2022-10-05
clients/js: always infer rt pub key from chainId

nhynes:
This PR works around an issue where wrapping an Ethers provider that in turn wraps Metamask prevents the library from detecting Metamask and falling back to fetching the runtime public key from the default gateway chosen by the `chainId` (Metamask does not like the custom `oasis_callDataPublicKey` RPC).


matevz (approved):
I can confirm this solves the `MetaMask - RPC Error: The method "oasis_callDataPublicKey" does not exist / is not available.` error in my case.

PR #72, opened 2022-10-12
clients/js: allow skipping ethers signer signed call

nhynes:


xmzheng:
It seems that we should also have some corresponding changes for handling unsigned calls in the runtime?

xmzheng (approved):

nhynes:
> It seems that we should also have some corresponding changes for handling unsigned calls in the runtime?

The runtime is okay with unsigned calls. They just have to be enveloped. [source](https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/modules/evm/src/lib.rs#L679)

PR #73, opened 2022-10-14
examples: Add hardhat-frontend example

matevz:
This PR adds a working hardhat example including the frontend part.

PR #76, opened 2022-10-19
clients/js: only generate lazy cipher once

nhynes:
Initializing the cipher wasn't sufficiently one-time. This PR fixes that and also #73. @pro-wh warned about this, but the fix wasn't good enough.

### DIFF
--- a/clients/js/src/cipher.ts
+++ b/clients/js/src/cipher.ts
@@ -235,19 +235,16 @@ export function lazy(generator: () => Promisable<Cipher>): Cipher {
   return new Proxy(
     {},
     {
-      get(target: { inner?: Cipher }, prop) {
+      get(target: { inner?: Promise<Cipher> }, prop) {
         // Props (Promiseable)
         if (prop === 'kind' || prop === 'publicKey') {
-          if (target.inner) return Reflect.get(target.inner, prop);
-          return Promise.resolve(generator()).then((inner: Cipher) => {
-            if (!target.inner) target.inner = inner;
-            return Reflect.get(target.inner, prop);
-          });
+          if (!target.inner) target.inner = Promise.resolve(generator());
+          return target.inner.then((c) => Reflect.get(c, prop));
         }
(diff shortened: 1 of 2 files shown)


clients/js/src/cipher.ts
```diff
@@ -235,19 +235,16 @@ export function lazy(generator: () => Promisable<Cipher>): Cipher {
...
-      get(target: { inner?: Cipher }, prop) {
+      get(target: { inner?: Promise<Cipher> }, prop) {
         // Props (Promiseable)
         if (prop === 'kind' || prop === 'publicKey') {
-          if (target.inner) return Reflect.get(target.inner, prop);
-          return Promise.resolve(generator()).then((inner: Cipher) => {
-            if (!target.inner) target.inner = inner;
-            return Reflect.get(target.inner, prop);
-          });
+          if (!target.inner) target.inner = Promise.resolve(generator());
```
pro-wh, line 241: there may be some misbehavior if `generator` causes us to re-enter this `get`

pro-wh (approved):
lg

btw did something actually go wrong with multiply generated ciphers? seems like it would only be a little extra work

clients/js/src/cipher.ts
nhynes: > btw did something actually go wrong with multiply generated ciphers

Matevž PR (#76) demonstrated an issue where sending two calls at once reliably causes the wrong cipher to be used. I think what was happening was that the `get` of the `Promise` properties only set the inner after a significant delay. The other branch looks like it runs exactly once.

> there may be some misbehavior if generator causes us to re-enter this get

```
let cipher;
let generator = () => new ReKeyCipher(cipher.key);
cipher = lazy(generator);
```
that's really weird. I hope that never happens but fully expect to @pro–wh again in a few months when it does.

matevz:
Thanks! I can confirm that this solves timeout and deoxys-ii errors in my PR.

PR #77, opened 2022-10-20
add Hardhat testnet chainId for local integration

xmzheng:
some corresponding changes to hookup with hardhat localhost network:

related PRs:
https://github.com/oasislabs/NomicFoundation-ethereumjs-vm/pull/1
https://github.com/oasislabs/NomicFoundation-hardhat/pull/1

### DIFF
--- a/clients/js/test/compat.spec.ts
+++ b/clients/js/test/compat.spec.ts
@@ -1,8 +1,6 @@
 import { parse as parseTx } from '@ethersproject/transactions';
 import * as cbor from 'cborg';
 import * as ethers from 'ethers';
-import nock from 'nock';
-import nacl from 'tweetnacl';
 
 import {
   wrap,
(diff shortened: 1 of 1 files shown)


clients/js/src/compat.ts
```diff
@@ -505,11 +507,20 @@ async function inferRuntimePublicKeySource(
...
+    if (
+      chainId == HARDHAT_LOCAL_NETWORK_CHAINID &&
+      'provider' in upstream &&
+      upstream['provider'] &&
+      'send' in upstream['provider']
+    ) {
+      return upstream['provider'];
+    } else {
+      return { chainId: chainId };
+    }
```
nhynes, line 523: This is a good idea, but it doesn't generalize since the hardhat network chain id can actually be set by the developer. How about this: try to get the public key using the provider, and if that fails fall back to the oasis web3 gateways.

xmzheng:
@nhynes please have a look at the refactor of always opportunistically using 'send' in the provider. If you agree on the framework, I will update/add corresponding unit tests.

clients/js/src/compat.ts
```diff
@@ -500,20 +499,39 @@ class EnvelopeError extends Error {}
...
   const isSigner = isEthersSigner(upstream);
+  let chainId: number;
   if (isSigner || isEthersProvider(upstream)) {
-    return {
-      chainId: isSigner
-        ? await upstream.getChainId()
-        : (await upstream.getNetwork()).chainId,
-    };
+    if (
+      'provider' in upstream &&
```
nhynes, line 509: This appears to only uses the web3 gateway when the upstream is an ethers signer. The way it was before, it would first check if it's ethers anything, and then pull specifically the provider out because that's all that's needed. Could you ensure that both (non-metamask-backed) ethers provider and signer can use this code path?

clients/js/src/compat.ts
```diff
@@ -500,20 +499,39 @@ class EnvelopeError extends Error {}
...
+    ) {
+      // first opportunistically try `send` from the provider
+      try {
+        const source = upstream['provider'] as {
+          send: (method: string, params: any[]) => Promise<any>;
+        };
+        const { key } = await source.send(OASIS_CALL_DATA_PUBLIC_KEY, []);
+        if (key) {
+          return arrayify(key);
+        }
```
nhynes, line 521:
```suggestion
        if (key) return arrayify(key);
```

clients/js/src/compat.ts
```diff
@@ -500,20 +499,39 @@ class EnvelopeError extends Error {}
...
+      } catch (e) {
```
nhynes, line 522:
```suggestion
      } catch {
```
to avoid the unused var

nhynes (commented):
This overall looks good. Could you please check at least the following cases

* fake metamask wrapped by ethers throws and reverts to chainId
* non-ethers upstream reverts to chainId
* non-metamask-backed ethers provider or signer uses the provider's RPC

You should be able to pass in some fakes during unit tests rather than doing anything with network mocks or browser testing.

nhynes (approved):

PR #84, opened 2022-11-14
clients/go: Add Golang compatibility library

### LINKED ISSUE #82: Golang compatibility library
Many users have requested a Golang version of the compatibility library. The beginnings of it already exist in this repo, so the remaining effort is to pack and the appropriate data in the Sapphire format.

Requirements:
* able to use the library to send Sapphire-format transactions through go-ethereum
* able to integrate with any version of go-ethereum
* compatible with Go [abigen](https://geth.ethereum.org/docs/dapp/native-bindings#generating-the-bindings)
    - wrapper type for `bind.ContractCaller`, `bind.ContractTransactor` that implements the same. this should be passed to `New` method of generated binding
    - the wrapper is just like the js one, but will use public methods for packing the data, so that the library can be used without an abi generator (raw go-ethereum)

aefhm:
#### Description
Resolve #82 and port JS compatibility functionality to Golang.

#### TODO
- [x] Add linting
- [x] Run tests automatically
- [x] Add README
- [x] Test import


clients/go/compat.go
```diff
@@ -0,0 +1,306 @@
...
+
+	blockHash := header.Hash()
+	leash := NewLeash(header.Nonce.Uint64(), header.Number.Uint64(), blockHash[:], DefaultBlockRange)
+
+	publicKey, _ := tweetnacl.ScalarMultBase(crypto.FromECDSA(&b.Key))
+	keypair := tweetnacl.KeyPair{
+		PublicKey: publicKey,
+		SecretKey: crypto.FromECDSA(&b.Key),
+	}
+	cipher, _ := NewX255919DeoxysIICipher(keypair, b.RuntimePublicKey)
```
nhynes, line 173: can you cache this in the wrapped backend? the cipher has a bit of setup cost that could be avoided.

clients/go/signed_calls.go
```diff
@@ -20,17 +24,39 @@ type Signer interface {
...
+	Body   []byte `json:"body"`
+	Format uint64 `json:"format,omitempty"` // reuse for now, TODO swap later
+}
+
+type PlainEnvelope struct {
+	Body   []byte `json:"body"`
+	Format uint64 `json:"format,omitempty"` // reuse for now, TODO swap later
+}
+
+type FancyEnvelope struct {
```
nhynes, line 51: lol

clients/go/signed_calls.go
```diff
@@ -50,13 +76,41 @@ func NewDataPack(signer Signer, chainId uint64, caller, callee []byte, gasLimit
...
+		BlockNumber: blockNumber,
+		BlockHash:   blockHash,
+		BlockRange:  blockRange,
+	}
+}
+
 func makeSignableCall(chainId uint64, caller, callee []byte, gasLimit uint64, gasPrice *big.Int, value *big.Int, data []byte, leash Leash) apitypes.TypedData {
+	toAddr := ZeroAddress
+	// callee should exist except for contract creation
+	if callee != nil {
```
nhynes, line 110: very good point

clients/go/compat.go
```diff
@@ -0,0 +1,306 @@
...
+		return 0, err
+	}
+
+	blockHash := header.Hash()
+	leash := NewLeash(header.Nonce.Uint64(), header.Number.Uint64(), blockHash[:], DefaultBlockRange)
+
+	dataPack, _ := NewDataPack(b.Signer, b.ChainID.Uint64(), call.From[:], addressToByte(call.To), DefaultGasLimit, call.GasPrice, call.Value, call.Data, leash)
+	call.Data = dataPack.Encode()
+
+	return b.Backend.EstimateGas(ctx, call)
```
nhynes, line 206: For this one, you're just going to have to return a constant big number, as sapphire doesn't support gas estimation at the moment. You can comment this part out, though. It'll be useful later.

nhynes (commented):
This is looking pretty good! Just a bit of refactoring and some "e2e" tests and it should be good to go.

clients/go/compat.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+			return tx.WithSignature(signer, signature)
+		},
+		Context: context.Background(),
+	}, nil
+}
+
+// NewKeyStoreTransactorWithChainID is a utility method to easily create a transaction signer from
+// an decrypted key from a keystore.
+func NewKeyStoreTransactorWithChainID(keystore *keystore.KeyStore, account accounts.Account, chainID *big.Int) (*bind.TransactOpts, error) {
```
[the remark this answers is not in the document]
aefhm: No. This is a comparable [wrapper](https://github.com/ethereum/go-ethereum/blob/c1aa1db69e74c71f251fc83cf7c120b4d0222728/accounts/abi/bind/auth.go#L105-L117) to the other methods.

Hmm, I think it may be possible to also remove these and code in the `GasPrice` and `GasLimit`.

clients/go/compat.go
```diff
@@ -0,0 +1,306 @@
...
+	baseTx := *types.NewTx(legacyTx)
+
+	signer := types.LatestSignerForChainID(tx.ChainId())
+	signature, _ := crypto.Sign(signer.Hash(&baseTx).Bytes(), &b.Key)
+	signedTx, _ := baseTx.WithSignature(signer, signature)
+
+	return b.Backend.SendTransaction(ctx, signedTx)
+}
+
+func (b WrappedBackend) CallContract(ctx context.Context, call ethereum.CallMsg, blockNumber *big.Int) ([]byte, error) {
```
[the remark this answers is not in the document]
aefhm: > This way we don't over-index on the generated method API and provide generality.

Makes sense.

> This function should just call the PackContractCall or whatever method, and then send it using the backend.

I see. Yeah, I like the idea of the more generalized solution. I guess the same applies as you did for the `SendTransaction`.

clients/go/signed_calls.go
aefhm: 😅 sanitized to be corporate okay

clients/go/compat.go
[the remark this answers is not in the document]
aefhm: Removed. Looks okay.

clients/go/compat.go
```diff
@@ -31,283 +35,241 @@ type NetworkParams struct {
...
+	txHash := signer.Hash(packedTx).Bytes()
+	signature, err := b.sign(*(*[32]byte)(txHash))
+	if err != nil {
+		return fmt.Errorf("failed to sign wrapped tx: %w", err)
+	}
+	signedTx, err := packedTx.WithSignature(signer, signature)
+	if err != nil {
+		return fmt.Errorf("failed to attach signature to wrapped tx: %w", err)
+	}
+	packedTxs.Delete(signedTx.Hash())
```
aefhm, line 263: Why do we have to delete this hash?

nhynes: to prevent memory use from growing without bound

Yawning (changes requested):

clients/go/compat.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/twystd/tweetnacl-go/tweetnacl"
+)
+
+const (
+	DefaultGasPrice   = 100_000_000_000
+	DefaultGasLimit   = 30_000_000 // Default gas params are assigned in the web3 gateway.
+	DefaultBlockRange = 15
+)
+
+// packedTxs stores the txs that have been packed in the signer and therefore do not need to be packed in `backend.SendTransaction`.
```
Yawning, line 24: Is the only reason why this exists, to support `PackTx` without a `WrappedBackend` instance?

clients/go/cipher.go
```diff
@@ -0,0 +1,291 @@
...
+	"fmt"
+	"math/rand"
+	"net/http"
+
+	"github.com/oasisprotocol/deoxysii"
+	"github.com/oasisprotocol/emerald-web3-gateway/rpc/oasis"
+	"github.com/oasisprotocol/oasis-core/go/common/cbor"
+	"github.com/oasisprotocol/oasis-core/go/common/crypto/mrae/api"
+	mrae "github.com/oasisprotocol/oasis-core/go/common/crypto/mrae/deoxysii"
+	"github.com/twystd/tweetnacl-go/tweetnacl"
```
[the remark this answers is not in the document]
Yawning: To be clear, my issue isn't with X25519 public/private keys being type-safe.  That's fine.  We shouldn't be pulling in tweetnacl of all things to do this...

clients/go/compat.go
nhynes: [This](https://github.com/ethereum/go-ethereum/blob/e5eb32acee19cc9fca6a03b10283b7484246b15a/accounts/abi/bind/base.go#L392) is the problem. The transaction returned by `transact` and therefore any of the abigenerated methods is the one that comes out of the signer. If the signer is not wrapped, you'll get a very confusing output where the tx hashes and data are all wrong. The solution this library implements is to create a wrapped signer that creates the packed transaction. Unfortunately, you can use the wrapped signer independently of a `WrappedBackend` (not sure why, but you can) or a `WrappedBackend` independently of a special signer (quite likely if the user is uninformed). That means that both the signer _and_ `WrappedBackend` must be capable of packing unpacked txs. The problem is that the backend doesn't know that it's receiving an already-packed tx from the signer (unless it decrypts the tx, but that's costly and slightly more brittle). The compromise solution is to just store txs that have likely been packed and send those along without modification.

clients/go/cipher.go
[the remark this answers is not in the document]
nhynes: Good point. It's gone now.

nhynes (approved):
Please make sure that this library is importable before merging!

aefhm:
I don't know how to have the pretty path of `"github.com/oasisprotocol/sapphire-paratime"` and to have the `go.mod` file in the root directory. I [see](https://github.com/golang/go/wiki/Modules#gomod) that:

```
The module path and the relative path from the go.mod to a package's directory together determine a package's import path.
```

So, my understanding is that keeping the `go.mod` file at root as suggested means the slightly longer [import path](https://github.com/oasislabs/wordleish/blob/c76008a280af4d9d81c2d4b639590602d986d068/cli/main.go#L21) which is probably okay.

nhynes:
> I don't know how to have the pretty path of `"github.com/oasisprotocol/sapphire-paratime"` and to have the `go.mod` file in the root directory. I [see](https://github.com/golang/go/wiki/Modules#gomod) that:

The go.mod is in the root because that's the easiest way to get _anything_ working. Based on my reading of the docs, you need to make a special tag that says the subdirectory. Let me see if I can get it working while you're sleeping.

PR #86, opened 2022-11-15
examples: Try to deploy without encrpyting the contract create tx

matevz:
Attempt to deploy contract in a way that contract verification can be made with the current blockscout explorer and possibly future tools.


matevz:
This was implemented in the [demo-starter hardhat task](https://github.com/oasisprotocol/demo-starter/blob/7cccc0be495906f8c0a9d13ba701ec94933d8ada/backend/hardhat.config.ts#L40-L53).

PR #87, opened 2022-11-20
Add more functionalities to cipher to reuse

xmzheng:


xmzheng:
I will add unit tests after we agree the format

clients/js/src/cipher.ts
```diff
@@ -80,6 +80,24 @@ export abstract class Cipher {
...
 
+  /** Decrypts the data contained within call */
+  public async decryptCallData(
+    nonce: Uint8Array,
+    ciphertext: Uint8Array,
+  ): Promise<Uint8Array> {
+    const { body: plainData } = cbor.decode(
+      await this.decrypt(nonce, ciphertext),
+    );
+    return plainData;
```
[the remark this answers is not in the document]
nhynes: In addition to testing and internal tools, this one I can imagine is useful for decoding what you've already sent by examining the chain.

clients/js/src/cipher.ts
```diff
@@ -80,6 +80,24 @@ export abstract class Cipher {
     return { data, nonce };
   }
 
+  /** Decrypts the data contained within call */
```
nhynes, line 83:
```suggestion
  /**
   * Decrypts the data contained within call.
   *
   * This is useful for creating tools, and also decoding previously-sent transactions that have used the same encryption key.
   */
```

clients/js/src/cipher.ts
```diff
@@ -80,6 +80,24 @@ export abstract class Cipher {
...
+    ciphertext: Uint8Array,
+  ): Promise<Uint8Array> {
+    const { body: plainData } = cbor.decode(
+      await this.decrypt(nonce, ciphertext),
+    );
+    return plainData;
+  }
+
+  /** Encrypts call result */
+  public async encryptCallResult(result: CallResult): Promise<Uint8Array> {
```
nhynes, line 95: This looks like it's just useful for testing and internal libraries, so I'm hesitant to expose this as the public interface of this library. Let's try something like
```suggestion
  /**
   * @hidden Encrypts a CallResult in the same way as would be returned by the runtime.
   * This method is not part of the SemVer interface and may be subject to change. 
   */
  public async encryptCallResult(result: CallResult): Promise<Uint8Array> {
```

nhynes (approved):
Looks good. I just had a few comments that you can apply before merging.

nhynes (approved):

PR #98, opened 2022-12-14
RNG example

matevz:
A simple example importing the `@oasisprotocol/sapphire-contracts/contracts/Sapphire.sol` wrapper and trying to generate two random numbers.


CedarMist:
This has been replaced by other examples which use the RNG.

PR #100, opened 2023-01-11
add cachedSignedCalls for re-use signatures for repeated signed calls

xmzheng:
add a cacheManager to signed_calls so that we can cache signatures for those repeated signed calls


clients/js/src/signed_calls.ts
```diff
@@ -21,8 +18,68 @@ const DEFAULT_GAS_LIMIT = 30_000_000;
...
+
+const _cacheManager = new SignedCallCacheManager();
+
+export function enableSignedCallCaches() {
+  _cacheManager.enable();
+}
+
+export function disableSignedCallCaches() {
+  _cacheManager.disable();
+}
```
nhynes, line 77:
```suggestion
```
Please don't add side effects, as it can confuse bundlers. It's better to initialize in wrap if needed.

clients/js/src/signed_calls.ts
```diff
@@ -21,8 +18,68 @@ const DEFAULT_GAS_LIMIT = 30_000_000;
 const DEFAULT_VALUE = 0;
 const DEFAULT_DATA = '0x';
 const zeroAddress = () => `0x${'0'.repeat(40)}`;
+const MAX_SAFE_BIG_NUM = 0x1fffffffffffff - 1;
 
-export type Signer = Pick<EthersSigner, 'getTransactionCount' | 'getChainId'> &
+class SignedCallCacheManager {
+  public enabled = true;
```
nhynes, line 24: Is this something that mainstream dapps will use? How hard would it be for the dapps that needs this to clear the cache when required, or make a subclass of this?

clients/js/src/signed_calls.ts
```diff
@@ -21,8 +18,68 @@ const DEFAULT_GAS_LIMIT = 30_000_000;
...
+    this.enabled = false;
+    cachedSignatures.clear();
+    cachedBlockNumbers.clear();
+  }
+
+  public setSignatureCache(
+    address: string,
+    hash: string,
+    signature: Uint8Array,
+  ) {
```
nhynes, line 45:
```suggestion
  public set(
    dataPack: SignedCallDataPack,
    signature: Uint8Array
  ) {
```
having the API like this frees you to use any method of digesting the data pack into a map key

clients/js/src/signed_calls.ts
```diff
@@ -21,8 +18,68 @@ const DEFAULT_GAS_LIMIT = 30_000_000;
...
+    if (!this.enabled) return;
+    if (!this.cachedSignatures.has(address))
+      this.cachedSignatures.set(address, new Map<string, Uint8Array>());
+    this.cachedSignatures.get(address)!.set(hash, signature);
+  }
+
+  public getSignatureCache(
+    address: string,
+    hash: string,
+  ): Uint8Array | undefined {
```
nhynes, line 55:
```suggestion
  public get(
    address: string,
    chainId: number,
    call: string,
  ): Uint8Array | undefined {
```
similarly

clients/js/src/signed_calls.ts
```diff
@@ -21,8 +18,68 @@ const DEFAULT_GAS_LIMIT = 30_000_000;
...
+  }
+
+  public setBlockNum(chainId: number, blockNum: BlockTag) {
+    if (!this.enabled) return;
+    this.cachedBlockNumbers.set(chainId, blockNum);
+  }
+
+  public getBlockNum(chainId: number): BlockTag | undefined {
+    return this.cachedBlockNumbers.get(chainId);
+  }
```
nhynes, line 66:
```suggestion
```
Instead of doing it this way, please store the actual leash with the cached signature, then before sending, check if it's still valid, if so then send, otherwise regenerate the signature. If you do it this way, you end up being mostly unable to use the `blockRange` part of the leash.

Here's the algorithm I would suggest
1. digest the non-leash part of the call using either `_hashTypedData` or canonicalized JSON
2. check if the signature+leash is in the cache
3. check if the leash is still valid. you'll likely want a function called `leashHasSlack` to encapsulate this functionality
4. if the leash is still valid, attach the signature to the data pack and send. if the leash is invalid, invalidate the cache entry and recurse on `signCall`

This method will requiring checking the leash before sending, but makes it easier to handle short or custom `blockRange` and `baseBlock`s, which are both important for security.

nhynes (changes requested):
This is a good start. I made some comments around making the cache algorithm more amenable to more secure values of `blockRange`.

nhynes (approved):

PR #101, opened 2023-01-12
change signed call base block number to be latest -2

xmzheng:


clients/js/src/signed_calls.ts
```diff
@@ -125,7 +125,7 @@ async function makeLeash(
   } else {
     if (signer._checkProvider) signer._checkProvider('getBlock');
     const latestBlock = await signer.provider!.getBlock('latest');
-    blockP = signer.provider!.getBlock(latestBlock.number - 1); // The latest block is not historical.
+    blockP = signer.provider!.getBlock(latestBlock.number - 2); // The latest block is not historical.
```
nhynes, line 128:
```suggestion
    blockP = signer.provider!.getBlock(latestBlock.number - 2); // The latest block is not historical and may not even be the latest light client verifiable block.
```

nhynes (approved):
Please update the comment. Otherwise looks good. Thanks for the fix.

xmzheng:
@nhynes please help publish to npm after merge.

PR #103, opened 2023-01-17
clients: Fix Sapphire Mainnet configuration

kostko:


matevz (approved):
When deployed to npm, I suppose we also need to bump [sapphire-hardhat](https://github.com/oasisprotocol/sapphire-paratime/blob/main/integrations/hardhat/package.json) for the new dependency change.

PR #112, opened 2023-02-06
unwrap the unsigned query for call format

xmzheng:
the res has callformat even for unsigned query, we need to unwrap it.


nhynes:
Thanks. Can you please add a test for this?

nhynes (approved):

PR #116, opened 2023-02-21
contracts: add curve25519 pk precompile

nhynes:


kostko (approved):

contracts/contracts/Sapphire.sol
```diff
@@ -59,17 +64,42 @@ library Sapphire {
...
+     * @return sk The Curve25519 secret key. Pairs well with {`deriveSymmetricKey`}.
+     */
+    function generateCurve25519KeyPair(bytes memory pers)
+        internal
+        view
+        returns (Curve25519PublicKey pk, Curve25519SecretKey sk)
+    {
+        bytes memory scalar = randomBytes(32, pers);
+        // Twiddle some bits, as per RFC 7748 §5.
+        scalar[0] &= 0xf8; // Make it a multiple of 8 to avoid small subgroup attacks.
```
HarryR, line 80: It is unnecessary to perform a scalar clamp here. The curve25519-dalek library does it when casting bytes32 to field element.

Via:
 * https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/modules/evm/src/precompile/confidential.rs#L112
 * https://github.com/dalek-cryptography/x25519-dalek/blob/main/src/x25519.rs#L253
 * https://github.com/dalek-cryptography/curve25519-dalek/blob/a63e14f4ded078d6bf262ba0b3f47026bdd7f7c0/src/scalar.rs#L300

contracts/contracts/Sapphire.sol
```diff
@@ -14,12 +14,17 @@ library Sapphire {
         0x0100000000000000000000000000000000000003;
     address private constant DECRYPT =
         0x0100000000000000000000000000000000000004;
-    address private constant GENERATE_KEYPAIR =
+    address private constant GENERAGE_SIGNING_KEYPAIR =
```
HarryR, line 17: Should be `GENERATE_SIGNING_KEYPAIR`

kostko: Indeed! Can you open up a PR with the fix? Thanks a lot!

PR #117, opened 2023-02-22
Add OPL SDK

nhynes:
This SDK makes it very easy to use OPL. All you need to do is split your dapp and drop in the addresses.

Features:
* minimal boilerplate code
* partial message decoding
* automatic mainnet/testnet switching
* specify the home chain using  name instead of id
* mostly gas efficient (except for one place, which has been commented)

Here's a hello world example:

```solidity
pragma solidity ^0.8.0;

contract ScoreNFT is Host {
    constructor(address _scorer) Host(_scorer) {
        registerEndpoint("score", _oplReceiveScore);
    }
    
    function mintScore() external {
        postMessage("genScore", msg.sender);
    }

    function _oplReceiveScore(bytes calldata args)  returns (Result) {
        (address whom, uint256 score) = abi.decode(args, (address, uint256));
        _mint(msg.sender, score);
        return Result.Success;
    }
}
```

```solidity
pragma solidity ^0.8.0;

contract Scorer is Enclave {
    constructor(address _nft) Enclave(_nft, autoswitch("ethereum")) {
        registerEndpoint("genScore", _oplGenScore);
    }

    function _oplGenScore(bytes calldata args)  returns (Result) {
        (address whom) = abi.decode(args, (address));
        postMessage("score", secretAlgorithm());
        return Result.Success;
    }
}
```

I'd also like to generate a lot of overloads for the receiver methods so you get automatic decoding, but that'll might make the compiler slow.


contracts/contracts/opl/Enclave.sol
```diff
@@ -0,0 +1,17 @@
...
+import "./Endpoint.sol";
+
+/**
+ * @title OPL Enclave
+ * @dev The Sapphire-side of an OPL dapp.
+ */
+contract Enclave is Endpoint, ERC2771Context {
+    constructor(address _host, bytes32 _hostChain)
+        Endpoint(_host, _hostChain)
+        ERC2771Context(block.chainid == 0x5aff ? address(0) : address(1))
```
nhynes, line 15: the Sapphire gas relayers would go here

contracts/contracts/opl/Endpoint.sol
```diff
@@ -0,0 +1,239 @@
...
+        if (_chainId == 0xfa)
+            // fantom
+            return (0xFF4E183a0Ceb4Fa98E63BbF8077B929c8E5A2bA4, false);
+        if (_chainId == 0xfa2)
+            // fantom testnet
+            return (0xb92d6933A024bcca9A21669a480C236Cbc973110, true);
+        if (_chainId == 0x505)
+            // moonriver
+            return (0x940dAAbA3F713abFabD79CdD991466fe698CBe54, false);
+        if (_chainId == 0x5aff)
```
kostko, line 180: No Mainnet yet?

contracts/contracts/opl/Endpoint.sol
```diff
@@ -0,0 +1,239 @@
...
+        if (name == "arbitrum-testnet") return 0x66eeb;
+        return 0;
+    }
+
+    function _getChainConfig(uint256 _chainId)
+        internal
+        pure
+        returns (address _bus, bool _isTestnet)
+    {
+        if (_chainId == 1)
```
kostko, line 153: Would be nice to put the source for these in a comment (I assume these are from https://im-docs.celer.network/developer/contract-addresses-and-rpc-info).

contracts/contracts/opl/Endpoint.sol
```diff
@@ -0,0 +1,239 @@
...
+        );
+        ++txSeq;
+    }
+
+    function executeMessage(
+        address _sender,
+        uint64 _senderChainId,
+        bytes calldata _message,
+        address
+    ) external payable returns (uint256) {
```
kostko, line 88: I assume we are avoiding Celer imports (e.g. for types, auth check modifiers) for a reason (overhead, flexibility)?

contracts/contracts/opl/Enclave.sol
kostko, line 15: What's with these trusted forwarder addresses? Is the plan to plug in the GSN here once available?

kostko (commented):
Nice!

The example above has the following in `_oplReceiveScore`:
```solidity
_mint(msg.sender, score);
```

Shouldn't it instead use `whom` not `msg.sender` as the sender would probably be the relayer?

contracts/contracts/opl/Endpoint.sol
nhynes: `GPL-3.0-or-later` :P

contracts/contracts/opl/Endpoint.sol
nhynes: Celer said they'd be deploying within a week.

kostko (approved):
Looks good, would be nice to have some E2E tests for this somewhere.

PR #121, opened 2023-03-18
Decode encrypted Ok and plain Fail

nhynes:
v0.5.0 introduced a change wherein the success or failure of the call is exposed: `Ok(Encrypt(Ok(result)))` and `Err(error)`. This PR adds decoding of encrypted `Ok`s to the client libraries.

### DIFF
--- a/clients/go/cipher.go
+++ b/clients/go/cipher.go
@@ -31,9 +31,9 @@ var (
 )
 
 type CallResult struct {
-	Fail    *Failure `json:"failure,omitempty"`
-	OK      []byte   `json:"ok,omitempty"`
-	Unknown *Unknown `json:"unknown,omitempty"`
+	Fail    *Failure      `json:"failure,omitempty"`
+	OK      []byte        `json:"ok,omitempty"`
+	Unknown *AeadEnvelope `json:"unknown,omitempty"`
 }
 
 type Inner struct {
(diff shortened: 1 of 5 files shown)


kostko (approved):

aefhm:
> v0.5.0 introduced

(nit) We mean v0.5.0 will introduce?

clients/go/cipher.go
```diff
@@ -213,35 +213,38 @@ func (c X25519DeoxysIICipher) DecryptCallResult(response []byte) ([]byte, error)
...
-	if callResult.Unknown != nil {
-		decrypted, err := c.Decrypt(callResult.Unknown.Nonce, callResult.Unknown.Data)
-		if err != nil {
-			return nil, err
+	var aeadEnvelope AeadEnvelope
+	if callResult.OK != nil {
+		if err := cbor.Unmarshal(callResult.OK, &aeadEnvelope); err != nil {
+			return callResult.OK, nil
 		}
+	} else if callResult.Unknown != nil {
```
aefhm, line 221: I thought this path wouldn't occur post `0.5.0`?

aefhm (approved):

kostko:
> > v0.5.0 introduced
> 
> (nit) We mean v0.5.0 will introduce?

There is already a [v0.5.0-testnet](https://github.com/oasisprotocol/sapphire-paratime/releases/tag/v0.5.0-testnet).

clients/go/cipher.go
kostko: The client should still work for previous versions.

PR #132, opened 2023-06-13
Return fee used by postMessage

### LINKED ISSUE #131: postMessage should return how much ROSE it would require to submit the cross-chain message
https://github.com/oasisprotocol/sapphire-paratime/blob/55284f3951859c035dcc6e99aaf228d4acfaa880/contracts/contracts/opl/Endpoint.sol#L90

Without returning gas transferred to Celer MessageBus an Endpoint inheriting contract doesn't know how much it cost to send a message, so would be unable to charge a different user appropriately and rely on some external benevolent entity to ensure the contract has enough to pay for the relaying.

This could also cause havok if the contract is relaying messages to other chains and also trying to maintain an internal ROSE balance of external user deposits, the contracts internal accounting could have a mismatch of actual ROSE it holds vs how much it thinks it holds.

A temporary workaround would be to duplicate this code before calling `postMessage`, except `txSeq` is private so inaccessible to inheriting contracts:

```solidity
        bytes memory envelope = abi.encodePacked(
            bytes4(keccak256(_method)),
            txSeq,
            _message
        );
        uint256 fee = estimateFee(envelope.length);
```

CedarMist:
Closes #131

### DIFF
--- a/contracts/contracts/opl/Endpoint.sol
+++ b/contracts/contracts/opl/Endpoint.sol
@@ -87,7 +87,10 @@ contract BaseEndpoint is Context {
         return postMessage(_method, "");
     }
 
-    function postMessage(bytes memory _method, bytes memory _message) internal returns (uint256) {
+    function postMessage(bytes memory _method, bytes memory _message)
+        internal
+        returns (uint256)
+    {
         bytes memory envelope = abi.encodePacked(
             bytes4(keccak256(_method)),
             txSeq,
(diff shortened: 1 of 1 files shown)


nhynes (approved):

contracts/contracts/opl/Endpoint.sol
```diff
@@ -87,7 +87,10 @@ contract BaseEndpoint is Context {
         return postMessage(_method, "");
     }
 
-    function postMessage(bytes memory _method, bytes memory _message) internal {
+    function postMessage(bytes memory _method, bytes memory _message)
```
nhynes, line 90:
```suggestion
    /// Calls the remote endpoint, returning the amount of native token charged for the operation.
    function postMessage(bytes memory _method, bytes memory _message)
```

PR #136, opened 2023-06-20
Use SigningAlg instead of uint256 in Sapphire.signDigest

CedarMist:
to be consistent with `generateSigningKeyPair` and `verifyDigestSignature`.


CedarMist:
This shouldn't be merged, identified that the Sapphire.sol doesn't match the implementation in https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/modules/evm/src/precompile/confidential.rs

Am making working tested example for the different signing algs, then will update Sapphire.sol and documentation to match.

PR #137, opened 2023-06-21
Fixes Sapphire.sol sign/verify & adds sapphire-dev unit tests with hardhat

CedarMist:
This pull request fixes Sapphire.sol signing & verifying - which were broken / non-functional.

 - [x] https://github.com/oasisprotocol/docs/pull/460
 - [x] https://github.com/oasisprotocol/sapphire-paratime/issues/138
 - [x] Fix linter errors
 - [x] Sync branch with sapphire-paratime#master
 - [x] Tests: verify modifying any parameter by 1 random bit causes verification to fail
 - [x] CI with sapphire-dev using github actions
 - [x] Basic unit tests for all signing algorithms
 - [x] Update version to 0.2.2

Previously the `Sapphire.signDigest` and `Sapphire.verifyDigestSignature` functions did not match the Sapphire implementations in [oasis-sdk - precompile/confidential.rs](https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/modules/evm/src/precompile/confidential.rs).

The EVM precompile accepts both `context_or_hash` and `message` as different signature algorithms use either or both. However, the Sapphire.sol library only had a `digest` parameter (equivalent to `context_or_hash`) and `message` parameter. Thus when the precompile was called it would not function correctly and no signatures generated would validate if you followed the examples in the docs.

This change adds the `contextOrHash` and `message` parameter, changing the function names to `sign` and `verify`, and provides unit tests (currently testing against sapphire 0.4.0, as the sapphire-dev container hasn't been updated to 0.5.2).

------

New solidity call:

```solidity
function sign(
        SigningAlg alg,
        bytes memory secretKey,
        bytes memory contextOrHash,
        bytes memory message
    ) internal view returns (bytes memory signature) {
        (bool success, bytes memory sig) = SIGN_DIGEST.staticcall(
            abi.encode(alg, secretKey, contextOrHash, message)
        );
```

[Existing rust code](https://github.com/oasisprotocol/oasis-sdk/blob/ca630e7d48986bd102d7aa477a48f8966a3d1d23/runtime-sdk/modules/evm/src/precompile/confidential.rs#L282)

```rust
pub(super) fn call_sign(handle: &mut impl PrecompileHandle) -> PrecompileResult {
    let mut call_args = ethabi::decode(
        &[
            ParamType::Uint(256), // signature type
            ParamType::Bytes,     // private key
            ParamType::Bytes,     // context or precomputed hash bytes
            ParamType::Bytes,     // message; should be zero-length if precomputed hash given
        ],
        handle.input(),
    )
```

----------

New solidity call:

```solidity
function verify(
        SigningAlg alg,
        bytes memory publicKey,
        bytes memory contextOrHash,
        bytes memory message,
        bytes memory signature
    ) internal view returns (bool verified) {
        (bool success, bytes memory v) = VERIFY_DIGEST.staticcall(
            abi.encode(alg, publicKey, contextOrHash, message, signature)
        );
```

[Existing rust code](https://github.com/oasisprotocol/oasis-sdk/blob/ca630e7d48986bd102d7aa477a48f8966a3d1d23/runtime-sdk/modules/evm/src/precompile/confidential.rs#LL338C1-L348C6)

```rust
pub(super) fn call_verify(handle: &mut impl PrecompileHandle) -> PrecompileResult {
    let mut call_args = ethabi::decode(
        &[
            ParamType::Uint(256), // signature type
            ParamType::Bytes,     // public key
            ParamType::Bytes,     // context or precomputed hash bytes
            ParamType::Bytes,     // message; should be zero-length if precomputed hash given
            ParamType::Bytes,     // signature
        ],
        handle.input(),
    )
```

----


CedarMist:
I would like to add an on-chain method to convert compressed Secp256k1 points returned from the key generation precompile into an Ethereum compatible address. This requires recovering y coordinate from the sign bit so it can be hashed. Example: https://github.com/witnet/elliptic-curve-solidity/blob/347547890840fd501809dfe0b855206407136ec0/contracts/EllipticCurve.sol#L113

Making Sapphire.sol `ecrecover` compatible will be very useful, but I will not include it in this pull request.

CedarMist:
Unsure why linter step is failing, the CI job doesn't output any errors just `[warn]` with no details, and when I run it locally I get:

```
$ pnpm run lint::prettier

> @oasisprotocol/sapphire-contracts@0.2.2 lint::prettier /src/contracts
> prettier --cache --check --plugin-search-dir=. --cache '*.json' '**/*.ts' '**/*.sol'

Checking formatting...
All matched files use Prettier code style!
```

.github/workflows/contracts-test.yaml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          path: ${{ steps.pnpm-cache.outputs.pnpm_cache_dir }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
+      - name: Install dependencies
+        run: pnpm --filter '@oasisprotocol/sapphire-*' install
+      - name: Build JS client
+        run: pnpm build
+        working-directory: clients/js
```
matevz, line 57: I'd put `working-directory` above the `run` command. Same below.

CedarMist: This has been modified

kostko (approved):
Nice! We should also do a new release of `@oasisprotocol/sapphire-contracts`.

jberci (approved):

CedarMist:
After merging to master one of the tests failed.

The previous two pre-merge runs succeeded:

 * https://github.com/oasisprotocol/sapphire-paratime/actions/runs/5348593749/jobs/9698659290
 * https://github.com/oasisprotocol/sapphire-paratime/actions/runs/5348593749/jobs/9698659290

```
  1) Precompiles
       Sign & Verify:

      AssertionError: expected true to equal false
      + expected - actual

      -true
      +false
      
      at testSignThenVerify (test/precompiles.ts:35:7)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async Context.<anonymous> (test/precompiles.ts:122:7)
```

https://github.com/oasisprotocol/sapphire-paratime/blob/9a74e57e72b06ba86ec8454062b8c0a5281edb97/contracts/test/precompiles.ts#L118-L123

https://github.com/oasisprotocol/sapphire-paratime/blob/9a74e57e72b06ba86ec8454062b8c0a5281edb97/contracts/test/precompiles.ts#L25-L36

The test is to verify if changing the message invalidates the signature when using Ed25519Pure. It's returning true when it should return false. Unsure why this would fail.

CedarMist:
Re-ran CI tests for main again... It works, main branch now has green checkmark.

PR #140, opened 2023-06-28
clients/js: wrap ethers6

nhynes:
This PR adds ethers6 support to the compat lib.

The compat lib is getting unmaintainable*, but it should be mostly fine for another few years until the ecosystem changes again. What I would recommend is to split this compat lib into a stub that auto-detects the kind of provider it's getting, and then dynamically import the specific wrapper for it.

\* The signed calls module over-exports, which makes it harder to retain backwards compatibility. The provider detection code needs to be centralized. The overall idea of hooks and `wrap` is working well, but tree-shakeability is becoming an issue, as ethers5 and ethers6 are both large codebases.


CedarMist:
@nhynes re: maintainability, what is your opinion on https://github.com/MetaMask/eth-json-rpc-middleware ?

Will give this a more thorough review later, am concerned about compatibility with latest versions of Truffle or Hardhat.

nhynes:
> what is your opinion on https://github.com/MetaMask/eth-json-rpc-middleware

I think it's the right kind of thing, but after a cursory inspection, I do not see any obvious place where it integrates with any existing provider like hardhat or truffle. Is your expectation that this will itself interoperate with those providers?

clients/js/package.json
```diff
@@ -44,28 +44,27 @@
...
+    "@ethersproject/transactions": "^5.7.0",
     "@types/jest": "^28.1.8",
     "@types/node": "^18.7.18",
     "@types/node-fetch": "^2.6.2",
     "@typescript-eslint/eslint-plugin": "^5.38.0",
     "@typescript-eslint/parser": "^5.38.0",
     "eslint": "^8.23.1",
     "eslint-config-prettier": "^8.5.0",
-    "ethers": "^5.7.1",
+    "ethers5": "npm:ethers@^5.7.2",
```
matevz, line 67: Do we really want to keep ethers5? One can simply use the previous version of `@oasisprotocol/sapphire-paratime`, no?

nhynes: > One can simply use the previous version of @oasisprotocol/sapphire-paratime, no?

yes, but I don't want to create a new major version of this library yet. the current interface still hasn't hit its breaking point. I suspect people will not enjoy the slightly larger bundle size this causes, but maybe nobody will notice. Let's hold off on a v2 until the consensus is in on what the replacement should be.

CedarMist:
> Is your expectation that this will itself interoperate with those providers?

Yup, seems like it handles a variety of different wrapping types, e.g. `sapphire.wrap` could use eth-json-rpc-middleware behind the scenes to use an existing provider as an upstream, so would work with the existing hardhat/truffle integration pattern.

nhynes:
@CedarMist can you either link to or write a quick function that demonstrates `wrap(s: ethers.Signer): ethers.Signer` using `eth-json-rpc-middleware`? You can stub out the actual hook implementations.

CedarMist:
> can you either link to or write a quick function that demonstrates `wrap(s: ethers.Signer): ethers.Signer` using `eth-json-rpc-middleware`

Have made a ticket https://github.com/oasisprotocol/sapphire-paratime/issues/141 - is on my mental todo list as I only found the project yesterday and still need to have a more in-depth look at it.

nhynes:
I heard there were issues with truffle? Maybe someone could try upgrading the versions of that by following this PR as a template. This code isn't actually bad, but it could use a bit of reorganization to include a centralized detection function and a zero-bundle-size ethers5->ethers6 upgrade function.

CedarMist:
> I heard there were issues with truffle? 

I've made a temporary non-optimal hacky fix in #142 which tests against several versions of Truffle.

The problem with wrapping providers it's always going to have to deal with the underlying implementation, and there are still fundamental differences in how Truffle (e.g. HDWalletProvider) vs Ethers (either 5 or 6), metamask and whatever the `window.ethereum` provider is which conflict with one another when it comes to `.send` and `.sendAsync` which seems to be the root of the problem.

A good test would be trying to move everything to `.request` which is defined by EIP-1193, and seeing what breaks?

PR #153, opened 2023-07-13
clients/js: fix fetchRuntimePublicKey with Hardhat + Truffle

### LINKED ISSUE #146: When using Hardhat, `fetchRuntimePublicKey` falls through to `fetchRuntimePublicKeyByChainId`
There is a problem when using CI or a multi-container environment, for example Hardhat with the `sapphire-dev` where both are in containers on the same Docker network, where both containers have separate 172.16.x IP addresses.

The `fetchRuntimePublicKey` function isn't using the configured upstream JSON-RPC provider to request the call data public key. This means it's trying to access 127.0.0.1:8545 which won't work - because the sapphire-dev container is on another 172.x IP.

This is causing `hardhat test` to fail, but it works when running the tests locally, or when running the tests on the CI host rather than inside a container.

While this is annoying for CI, the more general problem is the provider detection code in `fetchRuntimePublicKey` doesn't seem to be functioning 100%.

-----------------

One of the problems is there's duplicated logic to find the correct `.send` method inside `fetchRuntimePublicKey` because it's being called prior to wrapping the provider.

If it's moved until after the provider has been wrapped, then it can use the consistent EIP-1193 `.request` method.

CedarMist:
Fixes #146

### DIFF
--- a/clients/js/src/compat.ts
+++ b/clients/js/src/compat.ts
@@ -638,8 +638,8 @@ function defer<T>() {
 export async function fetchRuntimePublicKey(
   upstream: UpstreamProvider,
 ): Promise<Uint8Array> {
-  const isSigner = isEthers5Signer(upstream) || isEthers6Signer(upstream);
-  const provider = isSigner ? upstream['provider'] : upstream;
+  const isEthersSigner = isEthers5Signer(upstream) || isEthers6Signer(upstream);
+  const provider = isEthersSigner ? upstream['provider'] : upstream;
   if (provider && 'send' in provider) {
     // first opportunistically try `send` from the provider
     try {
@@ -647,11 +647,12 @@ export async function fetchRuntimePublicKey(
         send: (
           method: string | {method:string, params:any[]},
           params?: any[] | ((err: any, ok?: any) => void)
-        ) => Promise<any>;
+        ) => Promise<any> | void;
       };
 
+      // For Truffle, turn a callback into an synchronous call
       let deferred = defer<any>();
-      const cb1 = function (err: any, ok?: any) {
+      const truffle_callback = function (err: any, ok?: any) {
         if( ok ) {
           deferred.resolve!(ok.result);
         }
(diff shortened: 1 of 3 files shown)


CedarMist:
Still getting intermittent errors, this seems to only happen part of the time, the other times it succeeds fine.

```
UnhandledRejections detected
Promise {
  <rejected> TypeError: callback is not a function
      at /home/runner/work/sapphire-paratime/sapphire-paratime/node_modules/.pnpm/truffle@5.10.1/node_modules/truffle/build/webpack:/packages/provider/wrapper.js:166:1
      at /home/runner/work/sapphire-paratime/sapphire-paratime/clients/js/src/compat.ts:514:23
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5) {
    uncaught: true,
    multiple: [ [Error] ]
  }
```

clients/js/src/compat.ts
```diff
@@ -628,39 +641,72 @@ class EnvelopeError extends Error {}
...
+        }
+        deferred.reject!(err);
+        return;
+      };
+
+      let resp;
+      if (
+        !isEthersSigner &&
+        !isEthers5Provider(provider) &&
+        !isEthers6Provider(provider)
```
aefhm, line 670: (nit, optional) Maybe we could just say this is checking for Truffle more explicitly.

examples/truffle/test/metacoin.js
```diff
@@ -14,6 +14,8 @@ contract('MetaCoin', (accounts) => {
 
     assert.equal(metaCoinEthBalance, 2 * metaCoinBalance, 'Library function returned unexpected function, linkage may be broken');
   });
+  /*
+  TODO: re-enable after sapphire-dev container in CI has multi-account support
```
aefhm, line 18: I'm cool with this and prefer local testing for now.

aefhm:
> Still getting intermittent errors, this seems to only happen part of the time, the other times it succeeds fine.
> 
> ```
> UnhandledRejections detected
> Promise {
>   <rejected> TypeError: callback is not a function
>       at /home/runner/work/sapphire-paratime/sapphire-paratime/node_modules/.pnpm/truffle@5.10.1/node_modules/truffle/build/webpack:/packages/provider/wrapper.js:166:1
>       at /home/runner/work/sapphire-paratime/sapphire-paratime/clients/js/src/compat.ts:514:23
>       at runMicrotasks (<anonymous>)
>       at processTicksAndRejections (node:internal/process/task_queues:96:5) {
>     uncaught: true,
>     multiple: [ [Error] ]
>   }
> ```
> 
> Currently there seems to be a 30-50% failure rate when running the CI tests :(

I think that is good enough for a patch granted not regression triggering.

clients/js/src/compat.ts
CedarMist: I think there's somehow a third provider type which gets added into the mix which is causing the non-deterministic failures. But that needs another iteration of bug hunting.

Will have a think about another approach to improve reliability after this is merged.

aefhm (approved):

PR #154, opened 2023-07-16
contracts: Added EIP-155 signer and extended EthereumUtils

CedarMist:
This allows contracts to sign and encode Ethereum compatible EIP-155 transactions.

As a follow-up to https://github.com/oasisprotocol/demo-voting/pull/9 where we should make this code commonly available in sapphire-contracts library.

Truffle tests are still intermittent with `callback` related errors (weeiird).

This will require publishing an updated sapphire-contracts NPM repo.

RLPWriter contract is MIT licensed from elsewhere, with (multiple) attributions in the file.

TODO:

 - [x] Commit v0.2.3 sapphire-contracts version
 - [x] Add tests for EIP-155 oroborus

### DIFF
--- a/contracts/contracts/EIP155Signer.sol
+++ b/contracts/contracts/EIP155Signer.sol
@@ -53,15 +53,14 @@ library EIP155Signer {
         address pubkeyAddr,
         bytes32 secretKey
     ) internal view returns (SignatureRSV memory ret) {
-        bytes memory encoded = encodeSignedTx(rawTx, SignatureRSV({v: rawTx.chainId, r: 0, s: 0}));
+        bytes memory encoded = encodeSignedTx(
+            rawTx,
+            SignatureRSV({v: rawTx.chainId, r: 0, s: 0})
+        );
 
         bytes32 digest = keccak256(abi.encodePacked(encoded));
 
-        ret = EthereumUtils.sign(
-            pubkeyAddr,
-            secretKey,
-            digest
-        );
+        ret = EthereumUtils.sign(pubkeyAddr, secretKey, digest);
     }
 
     /**
@@ -75,7 +74,11 @@ library EIP155Signer {
         bytes32 secretKey,
         EthTx memory transaction
     ) internal view returns (bytes memory) {
-        SignatureRSV memory rsv = signRawTx(transaction, publicAddress, secretKey);
+        SignatureRSV memory rsv = signRawTx(
+            transaction,
+            publicAddress,
+            secretKey
+        );
         rsv.v = (rsv.v - 27) + (transaction.chainId * 2) + 35;
         return encodeSignedTx(transaction, rsv);
     }
--- a/contracts/contracts/EIP155Signer.sol
+++ b/contracts/contracts/EIP155Signer.sol
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.0;
 
 import {Sapphire} from "./Sapphire.sol";
-import {EthereumUtils} from "./EthereumUtils.sol";
+import {EthereumUtils, SignatureRSV} from "./EthereumUtils.sol";
 import {RLPWriter} from "./RLPWriter.sol";
 
 /**
@@ -19,12 +19,6 @@ library EIP155Signer {
         uint256 chainId;
     }
 
-    struct SignatureRSV {
-        bytes32 r;
-        bytes32 s;
-        uint8 v;
-    }
-
     /**
      * Encode a signed EIP-155 transaction
      * @param rawTx Transaction which was signed
@@ -32,7 +26,7 @@ library EIP155Signer {
      */
     function encodeSignedTx(EthTx memory rawTx, SignatureRSV memory rsv)
         internal
-        view
+        pure
         returns (bytes memory)
     {
         bytes[] memory b = new bytes[](9);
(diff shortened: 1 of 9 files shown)


contracts/contracts/EIP155Signer.sol
```diff
@@ -0,0 +1,99 @@
...
+        bytes[] memory a = new bytes[](9);
+        a[0] = RLPWriter.writeUint(rawTx.nonce);
+        a[1] = RLPWriter.writeUint(rawTx.gasPrice);
+        a[2] = RLPWriter.writeUint(rawTx.gasLimit);
+        a[3] = RLPWriter.writeAddress(rawTx.to);
+        a[4] = RLPWriter.writeUint(rawTx.value);
+        a[5] = RLPWriter.writeBytes(rawTx.data);
+        a[6] = RLPWriter.writeUint(rawTx.chainId);
+        a[7] = RLPWriter.writeUint(0);
+        a[8] = RLPWriter.writeUint(0);
```
nhynes, line 71: this looks like an invocation of `encodeSignedTx` with a zero signature

contracts/contracts/EthereumUtils.sol
```diff
@@ -203,24 +222,63 @@ library EthereumUtils {
...
         view
         returns (
-            address pubkey_addr,
-            bytes32 r,
-            bytes32 s,
-            uint8 v
+            address pubkeyAddr,
+            bytes32 sigR,
+            bytes32 sigS,
+            uint8 sigV
```
nhynes, line 228: This looks like a `SignatureRSV`. Was there a reason not to use it? If it's to avoid encoding the struct, I'd first check whether solc optimizes that out (e.g. using the `viaIR` flag) before writing hand-optimized code.

contracts/contracts/EthereumUtils.sol
```diff
@@ -203,24 +222,63 @@ library EthereumUtils {
...
+        address pubkeyAddr,
+        bytes32 secretKey,
+        bytes32 digest
+    )
+        internal
+        view
+        returns (
+            bytes32 sigR,
+            bytes32 sigS,
+            uint8 sigV
```
nhynes, line 248: similarly

contracts/contracts/RLPWriter.sol
```diff
@@ -0,0 +1,249 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @custom:attribution https://github.com/bakaoh/solidity-rlp-encode
```
nhynes, line 5: I would recommend importing directly from the git commit. If I'm auditing the code, I don't really want to spend time figuring out what changes you made. If I'm maintaining the code, I'd prefer for it not to exist; and if it does, I'd also want the original tests.

contracts/test/eip155.ts
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import { expect } from 'chai';
+import { ethers } from 'hardhat';
+import { EIP155Tests__factory } from '../typechain-types/factories/contracts/tests';
+import { StaticJsonRpcProvider } from '@ethersproject/providers';
```
nhynes, line 6: This dependency already exists as `ethers.providers.StaticJsonRpcProvider`, so you can safely remove the extra installed package.

contracts/test/eip155.ts
```diff
@@ -0,0 +1,40 @@
...
+
+describe('EIP-155', function () {
+  async function deploy() {
+    const factory = (await ethers.getContractFactory(
+      'EIP155Tests',
+    )) as EIP155Tests__factory;
+    const x = await factory.deploy({ value: ethers.utils.parseEther('1') });
+    return { x };
```
nhynes, line 14:
```suggestion
    return { testContract };
```
also, this appears to be a `beforeAll`, which I'd recommend using

contracts/test/eip155.ts
```diff
@@ -0,0 +1,40 @@
...
+      gasLimit: 250000,
+      to: x.address,
+      value: 0,
+      data: '0x',
+      chainId: 0,
+    };
+    const signedTx = await x.sign(txobj);
+
+    // Submit signed transaction via plain JSON-RPC provider (avoiding saphire.wrap)
+    const plain_provider = new StaticJsonRpcProvider(
```
nhynes, line 31:
```suggestion
    const plainProvider = new StaticJsonRpcProvider(
```
please try to use consistent casing conventions

contracts/test/eip155.ts
```diff
@@ -0,0 +1,40 @@
...
+      ethers.provider.connection,
+    );
+    let plain_resp = await plain_provider.sendTransaction(signedTx);
+    let receipt = await x.provider.waitForTransaction(plain_resp.hash);
```
nhynes, line 35: You should be able to do
```suggestion
    let receipt = await tx.wait();
```
I think

nhynes (approved):
I left a few comments that I expect will be resolved in some way before merging. Otherwise LGTM

contracts/contracts/RLPWriter.sol
CedarMist: I'm not sure how to cherry-pick an individual file from a range of commits from another repo.

contracts/test/eip155.ts
CedarMist: This doesn't work, had to use the other approach.

contracts/contracts/EIP155Signer.sol
CedarMist: That's a good catch, thanks for pointing out how to remove duplication.

contracts/contracts/RLPWriter.sol
CedarMist: I'm going to leave this as-is for now, as the code is very straightforward.

nhynes: https://remix-ide.readthedocs.io/en/latest/import.html#import-from-a-github-url

PR #166, opened 2023-08-22
contracts: initial support for sha512, sha512/256, secp256r1 & subcall precompiles

CedarMist:
Adds wrappers for the new Sapphire-specific precompiles:

 - `sha512`
 - `sha512_256`
 - secp256r1 (`Sapphire.SigningAlg.Secp256r1PrehashedSha256`)
 - subcall (`Sapphire.subcall`)


CedarMist:
The as prices in 0.6 have changed too, so I will update the documentation to reflect that with the additional precompiles.

contracts/contracts/Sapphire.sol
```diff
@@ -1,28 +1,37 @@
 // SPDX-License-Identifier: Apache-2.0
-pragma solidity ^0.8.8;
+pragma solidity ^0.8.0;
```
aefhm, line 2: We wanted to standardize to `0.8.0`?

contracts/hardhat.config.ts
```diff
@@ -1,6 +1,5 @@
+//import '@oasisprotocol/sapphire-hardhat';
```
aefhm, line 1: Leave it out then?

CedarMist: Ah, my bad, I missed this while I was testing and merged hastily.

We should add an e2e test which verifies signed queries are working correctly.

contracts/contracts/Sapphire.sol
CedarMist: I'm unaware of anything in earlier versions of Solidity 0.8 which would break the Sapphire contracts.

We should decide an explicit minimum version of Solidity to support, along with PNPM and NodeJS, and make it clear in the README and consistent across the files?

aefhm: ➕ we should agree to minimum version and pin everything.

PR #171, opened 2023-09-04
runtime: Bump Oasis SDK to 0.7.2 and prepare 0.6.2-testnet

kostko:


runtime/src/lib.rs
```diff
@@ -122,8 +122,8 @@ impl sdk::Runtime for Runtime {
         if is_testnet() {
             // Testnet.
             Some(TrustRoot {
-                height: 16955217,
-                hash: "310020a2fad60820610303c015a413bee45b579a34cc1a90706bcbab8b06bd37".into(),
+                height: 17155743,
+                hash: "098e590c6b5502882374418b488d8d798ec939ddf450786c8bff2741e3d8c6f5".into(),
```
nhynes, line 126: I can't find this on the oasis explorer for some reason, but I have confirmed that this is correct on [Oasisscan](https://testnet.oasisscan.com/blocks/17155743)

nhynes (approved):

Issue #106, opened 2023-01-25
Hardhat Signiture problem - state doesn't change


Formator:
I have a problem with hardhat testing contracts on the sapphire network. 
I can't change the state of the already deployed smart contract with a combination when to use or when not to use `sapphire.wrap`.

To demonstrate my issue I have [forked](https://github.com/Formator/sapphire-paratime/tree/hardhat_sig_issue) this repo and added some [changes and scripts](https://github.com/oasisprotocol/sapphire-paratime/compare/main...Formator:sapphire-paratime:hardhat_sig_issue).

Before running please set env `PRIVATE_KEY_TESTER` beside `PRIVATE_KEY`.

This is the console output that I'm getting from `\examples\hardhat-boilerplate\scripts\deploy-transfer.js`
``` console
Deploying the contracts with the account: 0x84373E739d2c50eE55EF8a3AF7238f77804f83bF
Account balance: 6417194500000000000
Token address: 0x034BA58e63898205dCCf2dc57578Bb8fb930Bf08
Fauceting tester account: 0xC88309139e5809f4E736D700bB8Aca31540C67d8
```

And this is the console output that I'm getting from `\examples\hardhat-boilerplate\scripts\test.js`
``` console
Token address: 0x034BA58e63898205dCCf2dc57578Bb8fb930Bf08
Testing the contracts with the account: 0xC88309139e5809f4E736D700bB8Aca31540C67d8
1 - Check if tester account has enough ether
Tester account balance: 3995157700000000000
2 - Check if tester account has at least 100 tokens if he has't, send them from deployer account
Tester account token balance before transfer: 10
Tester account token balance after transfer: 10
3 - Check if tester account can transfer 10 token to deployer account
Tester account token balance before transfer: 10
Tester account token balance after transfer: 10
```

As you can see in test script I couldn't change the state of the smart contract even from the deployer signer which has worked on deploy-transfer script. 

Same contract, and same test scenario works from the frontend example.

nhynes:
Thanks for the reproducible test case. I think the reason you're seeing this is because all of the transactions revert because an unmodified `hardhat node` is not able to decode the encrypted payloads understood by Sapphire. I think that I have a fix for you. Please allow me a few hours!

Formator:
@nhynes No problem I will wait for your fix, thank you!

nhynes:
[...]

PR #175, opened 2023-09-06
contracts: submit pre-signed transaction

### LINKED ISSUE #179: client-js: Encrypt plain transactions in sendTransaction()
In the onchain-signer and demo-voting examples, we use plain, unwrapped provider in order to call `sendTransaction()` of the on-chain generated and signed EIP-155 transaction, because the wrapper aborts with `Error: Un-enveloped data was passed to sendRawTransaction, which is likely incorrect. Is the dapp using the Sapphire compat lib correctly?`.

The downside of this is that **the transaction is not encrypted**, so we need to encrypt the inner call on the chain to get privacy. This is expensive and makes no sense since the client could encrypt it with the symmetric key when submitting it.

Note: encryption is not related to the fact that the tx was signed by someone else (=on-chain signer in our case), so this should be perfectly doable.

CedarMist:
Fixes #179:

    Error: Un-enveloped data was passed to sendRawTransaction, which is likely incorrect. Is the dapp using the Sapphire compat lib correctly?

I have added tests transaction envelope tests:

    ✔ Wrapper encrypts self-signed transaction calldata (4038ms)
    ✔ Other-Signed transaction submission via un-wrapped provider (4223ms)
    ✔ Other-Signed transaction submission via wrapped provider (4115ms)
    ✔ Self-Signed transaction submission via wrapped provider (4061ms)

This slightly modifies sapphire.wrap behavior so if it receives a signed but unencrypted transaction from an address which it can't sign for the transaction will pass through as-is.


CedarMist:
The `repackRawTx` function rejects things that it could easily handle.

And if you CBOR encode the envelope you end up with the contract receiving CBOR encoded data, rather than the calldata...

CedarMist:
This will fix #179

CedarMist:
There are a few scenarios we want to handle with `sendRawTransaction`:

 * User submits signed but unencrypted transaction
   * Wrap in encrypted envelope, no need to re-sign it, submit to RPC endpoint
 * User submits signed & encrypted transaction
   * Pass it verbatim to RPC endpoint
 * User submits unsigned unencrypted transaction
   * Sign it, encrypt it & submit to RPC endpoint  

However, currently the wrapper will raise an error if you submit a transaction signed by somebody else, it will try to re-sign the transaction.

CedarMist:
So... looks like the error I want to fix is fundamentally unfixable.

But, instead I added a test to make sure the EIP155 module can sign and submit transactions, and that transactions submitted by the wrapped provider are encrypted.

CedarMist:
So I can fix this error where we submit pre-signed transactions through the provider which is attached to a signer with a different address. But we can't automatically encrypt pre-signed transactions.

matevz:
> So I can fix this error where we submit pre-signed transactions through the provider which is attached to a signer with a different address. But we can't automatically encrypt pre-signed transactions.

Why not? Encryption with an ephemeral key derived from the runtime and the client key has nothing to do with who signed the tx.

CedarMist:
> Why not? Encryption with an ephemeral key derived from the runtime and the client key has nothing to do with who signed the tx.

The calldata is signed, modifying the calldata by wrapping it in an encrypted envelope would invalidate the existing signature.

When I use:

```
const [owner] = await ethers.getSigners();
const provider = testContract.provider;
const wallet = sapphire.wrap(owner.connect(provider));
```

The `repackRawTx` function encrypts the calldata then asks the signer to re-sign the modified transaction, however with the JSON-RPC Signer it's 

```
Error: cannot alter JSON-RPC Signer connection (operation="connect", code=UNSUPPORTED_OPERATION, version=providers/5.7.2)
```

If I wrap the `owner` directly, e.g. `const wallet = sapphire.wrap(owner);` I get:

```
Error: signing transactions is unsupported (operation="signTransaction", code=UNSUPPORTED_OPERATION, version=providers/5.7.2)
```

And only via creating a new `ethers.Wallet` directly and `connect()` it to the provider does it work because it can sign the transaction.

If this happens via MetaMask and the dApp signs a transaction without encrypting it then sends it through a wrapped signer it will try to encrypt it before submission.

I have slightly modified the behavior and added test cases so it you pass a transaction not signed by the signers address it will not attempt to encrypt it and pass it on. This makes submitting the gasless transactions easier because you don't need to setup a different unwrapped provider to submit it.

------------------------

CedarMist:
This may be relevant: https://github.com/oasisprotocol/sapphire-paratime/issues/106

clients/js/src/compat.ts
```diff
@@ -573,11 +573,13 @@ async function repackRawTx(
     );
   } catch (e) {
     if (e instanceof EnvelopeError) throw e;
-    if (globalThis?.process?.env?.NODE_ENV !== 'test') {
-      console.trace(REPACK_ERROR);
-    }
   }
   const tx = ethers6.Transaction.from(raw);
+  if (tx.isSigned() && (!signer || (await signer!.getAddress()) != tx.from!)) {
+    // us we are be unable to re-sign the encrypted tx, so must passthrough when
```
aefhm, line 579:
```suggestion
    // we are be unable to re-sign the encrypted tx, so must passthrough when
```

contracts/test/eip155.ts
```diff
@@ -36,8 +74,52 @@ describe('EIP-155', function () {
...
+    // Calldata should be encrypted when we wrap the wallet provider
+    let x = await provider.sendTransaction(signedTx);
+    expect(entropy(x.data)).gte(3.8);
+    expect(x.data).not.eq(calldata);
+
+    let r = await provider.waitForTransaction(x.hash);
+    expect(r.logs[0].data).equal(EXPECTED_EVENT);
   });
+
+  // TODO: test error conditions?
```
aefhm, line 124: 👍

CedarMist: This is being handled in another PR: https://github.com/oasisprotocol/sapphire-paratime

I'm going to leave the TODO: in for now

contracts/test/eip155.ts
```diff
@@ -1,32 +1,73 @@
...
+  });
+
+  it('Wrapper encrypts self-signed transaction calldata', async function () {
+    const tx = await testContract.example();
+    expect(entropy(tx.data)).gte(EXPECTED_ENTROPY_ENCRYPTED);
+    expect(tx.data).not.eq(calldata);
+    expect(tx.data.length).eq(218);
+  });
+
+  /// Verifies that contract can sign a transaction, and we submit it via unwrapped provider
```
aefhm, line 57:
```suggestion
  /// Verify that contracts can sign transactions for submission with an unwrapped provider
```

contracts/test/eip155.ts
```diff
@@ -36,8 +77,53 @@ describe('EIP-155', function () {
...
       plainResp.hash,
     );
-    expect(receipt.logs[0].data).equal(
-      '0xfedcba9876543210fedcba9876543210fedcba9876543210fedcba9876543210',
-    );
+    expect(plainResp.data).eq(calldata);
+    expect(receipt.logs[0].data).equal(EXPECTED_EVENT);
+  });
+
+  /// Verifies that a contract can sign a transaction, and we can submit it via a wrapped provider
```
aefhm, line 84:
```suggestion
  /// Verify that contracts can sign transactions for submission with a wrapped provider
```

contracts/test/eip155.ts
```diff
@@ -36,8 +77,53 @@ describe('EIP-155', function () {
...
+  /// This lets transactions signed by other accounts pass-through the wrapped provider
```
aefhm, line 85:
```suggestion
  /// Transactions signed by other accounts should pass-through the wrapped provider
```

contracts/test/eip155.ts
```diff
@@ -36,8 +77,53 @@ describe('EIP-155', function () {
...
+      chainId: 0,
+    });
+
+    let plainResp = await provider.sendTransaction(signedTx);
+    let receipt = await provider.waitForTransaction(plainResp.hash);
+    expect(plainResp.data).eq(calldata);
+    expect(receipt.logs[0].data).equal(EXPECTED_EVENT);
+  });
+
+  /// Verifies that submitting a manually signed transaction will be encrypted
```
aefhm, line 106:
```suggestion
  /// Verify that the wrapped wallet will encrypt a manually signed transaction
```

aefhm (approved):
Non-blocking comments.

PR #177, opened 2023-09-08
examples: Add onchain-signer

matevz:
Adds a very simple onchain signer example primarily used in the docs and regression test:
- adds example/onchain-signer
- adds it to CI using sapphire-dev
- bumps pnpm to 8.x

Requires https://github.com/oasisprotocol/oasis-web3-gateway/issues/430 to be fixed (using 2nd account in the test-mnemonic as a signer).


matevz:
> Can we upgrade NodeJS to 18.
> 
> Currently it's using 16, which is now EOL as of 2023-09-11: https://nodejs.dev/en/about/releases/
> 
> Node 18 goes into LTS maintenance mode in October 2023, and will be supported until 2025-04-30

Nock introduced some breaking changes in v18. Let's do this in a separate PR: https://github.com/oasisprotocol/sapphire-paratime/pull/183

CedarMist (approved):

PR #182, opened 2023-09-13
integrations/hardhat: remove oasislabs/hardhat fork

CedarMist:
There was an issue where cipher.ts was calling the CJS compiled version of sapphire-paratime pulled from npmjs by the oasislabs/ethereumjs-vm package (a dependency of oasislabs/hardhat).


nhynes (approved):
The unwritten consequence of this is that the (mostly non-working) hardhat-node integration is now going away. That's fine, but I hope that a native sapphire-node will appear soon!

CedarMist:
This PR updates the version of sapphire-hardhat to '2.17.3' (in-line with the latest hardhat version) - although it now much more widely compatible with all hardhat 2.x versions.

After merging, the updated version will need to be scheduled to be published on NPMJS.com

PR #185, opened 2023-09-14
ci-test: Do not run all truffle combinations for onchain-signer

matevz:
This PR removes running onchain-signer for each truffle version combination, because it doesn't rely on truffle.

Followup to https://github.com/oasisprotocol/sapphire-paratime/pull/177.


CedarMist (approved):
Cool, didn't realize you could make it simpler like that.

PR #187, opened 2023-09-15
contracts: Add solidity API reference

matevz:
Adds Solidity API reference using `forge doc --build` to generate the markdowns and final HTMLs.

Depends on #186.

[Preview](https://api.docs.oasis.io/sol/sapphire-contracts)

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -4,6 +4,7 @@ on:
   push:
     paths:
       - clients/js/**
+      - contracts/**
     branches:
       - main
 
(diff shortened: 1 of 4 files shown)


CedarMist:
To reduce the size of the documentation many of the 'internal' items can be safely made 'private' without changing the consumable API.

I tried the multi-page `files` setting, but everything was titled 'Solidity API' and I'm unable to group it into modules and the title is hard-coded in their template. We could make our own modified template groups things into modules.

I found another documentation generator called [Dodoc](https://github.com/primitivefinance/primitive-dodoc) which has a better structure, e.g.: https://library.primitive.xyz/technical/smart-contracts/autogenerated-docs/manager/base/CashManager

I suggest removing tests from the docs:

```patch
diff --git a/contracts/hardhat.config.ts b/contracts/hardhat.config.ts
index 6bc37c2..94bb7c3 100644
--- a/contracts/hardhat.config.ts
+++ b/contracts/hardhat.config.ts
@@ -105,6 +105,14 @@ const config: HardhatUserConfig = {
   },
   docgen: {
     pages: 'single',
+    exclude: [
+      'tests/SigningTests.sol',
+      'tests/SubcallTests.sol',
+      'tests/SignedQueriesTests.sol',
+      'tests/EIP155Tests.sol',
+      'tests/HashTests.sol',
+      'OPL.sol'
+    ]
   },
 };
```

nhynes (approved):

nhynes:
OPL.sol is just a meta package https://github.com/oasisprotocol/sapphire-paratime/blob/main/contracts/contracts/OPL.sol

PR #173, opened 2023-09-04
Utilize gas estimation in client libraries

aefhm:
## Why
Relates to #39 

### Go
This appears fairly straight forward. I can test with a base Go script against our Vigil [contract](https://github.com/oasisprotocol/sapphire-paratime/blob/4e0434db8677862f9f1ca78735760b283148114e/examples/hardhat/contracts/Vigil.sol).

```Golang
package main

import (
	"context"
	"crypto/ecdsa"
	"fmt"
	"log"
	"math/big"

	"github.com/aefhm/vigil" // or however you do local modules
	"github.com/ethereum/go-ethereum/accounts/abi/bind"
	"github.com/ethereum/go-ethereum/crypto"
	"github.com/ethereum/go-ethereum/ethclient"
	sapphire "github.com/oasisprotocol/sapphire-paratime/clients/go"
)

var SapphireChainID = big.NewInt(0x5aff)

func main() {
	c, err := ethclient.Dial("https://testnet.sapphire.oasis.dev")
	if err != nil {
		log.Fatal(err)
	}

	privateKey, err := crypto.HexToECDSA("")
	if err != nil {
		log.Fatal(err)
	}

	publicKey := privateKey.Public()
	publicKeyECDSA, ok := publicKey.(*ecdsa.PublicKey)
	if !ok {
		log.Fatal("error casting public key to ECDSA")
	}

	var client *sapphire.WrappedBackend
	client, err = sapphire.WrapClient(*c, func(digest [32]byte) ([]byte, error) { return crypto.Sign(digest[:], privateKey) })

	if err != nil {
		log.Fatal(err)
	}

	fromAddress := crypto.PubkeyToAddress(*publicKeyECDSA)
	nonce, err := client.PendingNonceAt(context.Background(), fromAddress)

	gasPrice, err := client.SuggestGasPrice(context.Background())
	if err != nil {
		log.Fatal(err)
	}

	auth, _ := bind.NewKeyedTransactorWithChainID(privateKey, SapphireChainID)
	auth.Nonce = big.NewInt(int64(nonce))
	auth.Value = big.NewInt(0) // in wei
	auth.GasLimit = uint64(0)  // in units
	auth.GasPrice = gasPrice

	if err != nil {
		log.Fatal(err)
	}

	address, tx, _, err := vigil.DeployVigil(auth, client)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(address.Hex())
	fmt.Println(tx.Hash().Hex())


        // create secret
[...]

PR #191, opened 2023-09-25
Remove Truffle support from TS client

aefhm:
## Description
Consensys is [sunsetting](https://consensys.io/blog/consensys-announces-the-sunset-of-truffle-and-ganache-and-new-hardhat) Truffle.

Unblock https://github.com/oasisprotocol/sapphire-paratime/pull/173 and dependent on  https://github.com/oasisprotocol/docs/pull/547


matevz (changes requested):
Also delete examples/truffle.

examples/truffle/package.json
```diff
@@ -13,7 +13,7 @@
     "@truffle/hdwallet-provider": "^2.0.14"
   },
   "dependencies": {
-    "@oasisprotocol/sapphire-paratime": "workspace:^",
+    "@oasisprotocol/sapphire-paratime": "~1.1.1",
```
nhynes, line 16: we're pinning support for truffle to this version now? why not keep it 1.x, as we'll need to cut a 2.x at some point

nhynes (approved):

examples/truffle/package.json
aefhm: We plan on a patch release `1.1.2` with #175 and various other QOL issues.

This PR #191 will land in some `2.0.0` or `1.2.0` depending on how we like to name things.

matevz:
https://github.com/oasisprotocol/docs/pull/547 should be merged before truffle removal.

aefhm:
I think this PR is ready for next major release. 🙏 

cc @matevz

clients/js/src/compat.ts
```diff
@@ -653,42 +640,7 @@ export async function fetchRuntimePublicKey(
         ) => Promise<any> | void;
       };
 
-      // For Truffle, turn a callback into an synchronous call
```
matevz, line 656: Did you test the removal of this with Brave and Metamask? EIP-1193 is "Ethereum Provider JavaScript API", so this is not just truffle.

matevz (changes requested):

clients/js/src/compat.ts
aefhm: Test flow of using Metamask on Brave with our example through the Hardhat [boilerplate](https://docs.oasis.io/dapp/sapphire/browser/) right?

That flow does work, but now I'm suspicious of my local dev and testing. Good callout. Lemme keep digging. 🤔

aefhm: K. I see the error now. We had another fallback mechanism, but I think the client should be EIP-1193 compliant as you corrected! 👍

matevz (approved):
I tested demo-voting with MetaMask using the new client and everything seems to work. Good job.

PR #193, opened 2023-09-26
Skip subcall delegation test

aefhm:
## Description

Latest SDK usage https://github.com/oasisprotocol/sapphire-paratime/pull/192 prevents `0` delegation amount.

https://github.com/oasisprotocol/sapphire-paratime/blob/3a85e42e6c1cc090c28a521cf7df6353aa8a30c8/runtime/src/lib.rs#L196

### DIFF
--- a/contracts/test/subcall.ts
+++ b/contracts/test/subcall.ts
@@ -123,7 +123,8 @@ describe('Subcall', () => {
     expect(await contract.provider.getBalance(contract.address)).eq(0);
   });
 
-  it('consensus.Delegate', async () => {
+  // TODO: re-enable when delegation succeeds with value greater than 0
+  it.skip('consensus.Delegate', async () => {
     // Ensure contract has an initial balance.
     const initialBalance = parseEther('1.0');
     await ensureBalance(contract, initialBalance, owner);
--- a/contracts/test/subcall.ts
+++ b/contracts/test/subcall.ts
@@ -5,7 +5,7 @@ import * as cborg from 'cborg';
 import { SubcallTests } from '../typechain-types/contracts/tests/SubcallTests';
 import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address';
 import { parseEther } from 'ethers/lib/utils';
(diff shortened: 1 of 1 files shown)


kostko:
This is because in the consensus layer (at least for Mainnet and Testnet) such delegations would fail anyway. This just makes them fail sooner (e.g. immediately during the subcall instead of async in the next block). We can disable this for localnet if needed, but having this in is more realistic.

aefhm:
> This is because in the consensus layer (at least for Mainnet and Testnet) such delegations would fail anyway. This just makes them fail sooner (e.g. immediately during the subcall instead of async in the next block). We can disable this for localnet if needed, but having this in is more realistic.

+1 for more realism

> BTW, it seems unfortunate that in PRs these tests do not run against the updated runtime but instead run against the previous version. This prevents catching such things early.

Agree that is irksome.

> Is there a reason why delegating 100 consensus tokens (instead of zero) is too much for tests?

We can repro the failure but haven't pulled this one out yet.

aefhm:
Relates to https://github.com/oasisprotocol/sapphire-paratime/issues/195

aefhm:
> BTW, it seems unfortunate that in PRs these tests do not run against the updated runtime but instead run against the previous version. This prevents catching such things early.

The circular dependency probably requires future untangling? We just don't see a `runtime` dependency bump regularly except in this situation, so pulling in the Docker [image](https://github.com/oasisprotocol/sapphire-paratime/blob/3a85e42e6c1cc090c28a521cf7df6353aa8a30c8/.github/workflows/contracts-test.yaml#L13-L14) for integration testing make sense to me. We do have the post merge [run](https://github.com/oasisprotocol/sapphire-paratime/actions/runs/6311671067/job/17136138432) which can still tell us if things are broken before anyone makes a new release. 😬 

https://github.com/oasisprotocol/sapphire-paratime/blob/3a85e42e6c1cc090c28a521cf7df6353aa8a30c8/.github/workflows/contracts-test.yaml#L13-L14

kostko (approved):

PR #203, opened 2023-10-12
Add publish workflow

aefhm:
#### Description

Automatically publish newer versions of different JS packages to NPM if they are not present on the registry already on all pushes to the `main` branch.

#### TODO
- [x] Potentially update documentation to reflect the need for a post-merge tag step

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,26 @@
+name: publish
+
+on:
+  push:
+    tags:
+      - clients/js/[0-9].[0-9]+.[0-9]+
+      - contracts/[0-9].[0-9]+.[0-9]+
+      - integrations/hardhat/[0-9].[0-9]+.[0-9]+
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "18"
+      - name: Extract package from tag
+        id: extract-tag
+        run: |
+          echo "NPM_PACKAGE=$(echo ${{ github.ref_name }} | grep -oE '(clients/js|contracts|integrations/hardhat)')" >> $GITHUB_OUTPUT
+      - name: Publish ${{ github.ref_name }} to NPM
+        uses: JS-DevTools/npm-publish@v2
+        with:
+          token: ${{ secrets.NPM_TOKEN }}
+          package: ${{ steps.extract-tag.outputs.NPM_PACKAGE }}
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -3,9 +3,9 @@ name: publish
 on:
   push:
     tags:
-      - clients/js/[0-9].[0-9]+.[0-9]+
-      - contracts/[0-9].[0-9]+.[0-9]+
-      - integrations/hardhat/[0-9].[0-9]+.[0-9]+
+      - clients/js/v[0-9].[0-9]+.[0-9]+
+      - contracts/v[0-9].[0-9]+.[0-9]+
+      - integrations/hardhat/v[0-9].[0-9]+.[0-9]+
 
 jobs:
   publish:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,9 +65,9 @@ if the issue is not reported yet, [file a new one].
 
 ## Development
 
-### Building and Testing
+### Building
 
-Building and testing are documented in our
+Building documentation lies at the top-level
 [README](https://github.com/oasisprotocol/sapphire-paratime/blob/main/README.md).
 
 ### Contributing Code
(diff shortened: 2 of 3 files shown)


kostko:
Task linked: [CU-863h6mt6x Use CI to publish NPM packages on sapphire-paratime repo](https://app.clickup.com/t/863h6mt6x)

kostko:
Task linked: [CU-863h6mt6x Use CI to publish NPM packages on sapphire-paratime repo](https://app.clickup.com/t/863h6mt6x)

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,22 @@
+name: publish
+
+on:
+  push:
+    branches: main
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    strategy:
```
aefhm, line 10: This is slightly less efficient than running additional steps but uses less configuration and makes the publishing steps more higher level. I can go either way.

aefhm:
@CedarMist @matevz One alternative is something with more full-fledged change logs such as [changesets](https://github.com/changesets/changesets) which [`pnpm`](https://pnpm.io/workspaces#release-workflow) recommends. We don't have such a large monorepo with just 3 packages and mostly a linear development cycle, so I think something lightweight such as this GitHub [action](https://github.com/marketplace/actions/npm-publish) is sufficient.

CedarMist:
The simple github action looks easiest.

CedarMist:
@aefhm 

I don't think doing releases on every main merge will work.

Instead, we need to tag specific releases, and only upon tagging will that specific package get published.

kostko:
Yeah it should be based on tags, just note that in this repo we are also using tags to automatically release the runtime, so make sure to namespace the tags.

aefhm:
> @aefhm
> 
> I don't think doing releases on every main merge will work.
> 
> Instead, we need to tag specific releases, and only upon tagging will that specific package get published.

Just clarifying that this specific GitHub action (which would run on every `main` `push`) checks to see if the present package version has been published to NPM, and only publishes if not present.

Can definitely switch workflow trigger to a regex based on some name-spaced tag like (Core does [`go/v0.2202.10`](https://github.com/oasisprotocol/oasis-core/releases/tag/go%2Fv0.2202.10)), so perhaps `clients/js/{{version}}` and so on.

aefhm:
> Yeah it should be based on tags, just note that in this repo we are also using tags to automatically release the runtime, so make sure to namespace the tags.

Confirming that by this we mean Git tags? I think/agree we should reserve the use of the release release for the paratime itself. Clients can just live with a basic tag.

CedarMist (approved):
Always a bit nervous about testing NPM deploys, does that mean the final test is essentially in production?

aefhm:
> Always a bit nervous about testing NPM deploys, does that mean the final test is essentially in production?

More or less 😬. I couldn't test the `tag` piece (due to GitHub permissions), but we could always take down an NPM package if need be.

The one nice thing about this particular GitHub action versus writing more bare bones script is the release check, and we could always tag a patch first.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,26 @@
+name: publish
+
+on:
+  push:
+    tags:
+      - clients/js/[0-9].[0-9]+.[0-9]+
+      - contracts/[0-9].[0-9]+.[0-9]+
+      - integrations/hardhat/[0-9].[0-9]+.[0-9]+
```
matevz, line 8: I think we traditionally put the leading `v` in front of the version numbers in oasis-core, oasis-sdk, cli and also sapphire-paratime repo, for example:
```
oa@matevz-oa:~/oasis-sdk$ git tag
client-sdk/go/v0.1.0
client-sdk/go/v0.2.0
client-sdk/go/v0.3.0
client-sdk/go/v0.4.0
client-sdk/go/v0.5.0
client-sdk/go/v0.5.1
client-sdk/go/v0.6.0
client-sdk/go/v0.6.1
contract-sdk/v0.3.0
runtime-sdk/v0.1.0
runtime-sdk/v0.2.0
runtime-sdk/v0.3.0
runtime-sdk/v0.4.0
runtime-sdk/v0.5.0
runtime-sdk/v0.6.0
runtime-sdk/v0.7.0
runtime-sdk/v0.7.1
runtime-sdk/v0.7.2
runtime-sdk/v0.7.3
```

Perhaps it would be a good idea to have this consistent.

```suggestion
      - clients/js/v[0-9].[0-9]+.[0-9]+
      - contracts/v[0-9].[0-9]+.[0-9]+
      - integrations/hardhat/v[0-9].[0-9]+.[0-9]+
```

matevz (approved):
Small nit regarding the leading `v`. Otherwise, I look forward to having this in oasis-sdk repo for the `client-sdk/ts-web` as well! ;)

.github/workflows/publish.yaml
aefhm: Alright alright. Let's go for it moving forward 😄 

We only have a few old tags [here](https://github.com/oasisprotocol/sapphire-paratime/releases/tag/clients%2Fjs%2F1.1.1) and [there](https://github.com/oasisprotocol/sapphire-paratime/releases/tag/clients%2Fgo%2Fv0.9.1).

PR #204, opened 2023-10-16
README: added api-docs links

CedarMist:
Includes links to the API documentation in the README


matevz (approved):
Since p4w3 is over, you can also safely remove the `## Build` section at the end.

CedarMist:
Xi changed the p4w3 footer in https://github.com/oasisprotocol/sapphire-paratime/pull/203

PR #205, opened 2023-10-16
contracts: add gaspad and gasused precompiles

CedarMist:
Unfortunately this has different semantics than was initially intended as the initial gas cost for the calldata is added after the transaction is executed.

This means that there will be a 12 gas difference, even after padding, between every NULL and non-NULL byte of the calldata. While this has a very small impact and won't reveal too much sensitive information, it will still reveal *some* information about the encrypted calldata.

A suggested workaround is to add `bytes calldata fuz` argument and filling it with one non-zero byte for every zero byte in the other arguments.

### DIFF
--- a/contracts/contracts/Sapphire.sol
+++ b/contracts/contracts/Sapphire.sol
@@ -23,6 +23,10 @@ library Sapphire {
         0x0100000000000000000000000000000000000007;
     address internal constant CURVE25519_PUBLIC_KEY =
         0x0100000000000000000000000000000000000008;
+    address internal constant GAS_USED =
+        0x0100000000000000000000000000000000000009;
+    address internal constant GAS_PAD =
+        0x010000000000000000000000000000000000000a;
 
     // Oasis-specific, general precompiles
     address internal constant SHA512_256 =
@@ -224,6 +228,32 @@ library Sapphire {
         );
         require(success, "verify: failed");
         return abi.decode(v, (bool));
+    }
+
+    /**
+     * @dev Set the current transactions gas usage to a specific amount
+     * @param toAmount Gas usage will be set to this amount
+     * @custom:see @oasisprotocol/oasis-sdk :: precompile/gas.rs :: call_pad_gas
+     *
+     * Will cause a reversion if the current usage is more than the amount
+     */
+    function gaspad(uint128 toAmount) internal view
+    {
+        (bool success,) = GAS_PAD.staticcall(
+            abi.encode(toAmount)
+        );
+        require(success, "verify: failed");
+    }
+
+    /**
+     * @dev Returns the amount of gas currently used by the transaction
+     * @custom:see @oasisprotocol/oasis-sdk :: precompile/gas.rs :: call_gas_used
+     */
+    function gasused() internal view returns (uint64)
+    {
+        (bool success, bytes memory v) = GAS_USED.staticcall("");
+        require(success, "gasused: failed");
+        return abi.decode(v, (uint64));
     }
 }
 
--- a/contracts/contracts/Sapphire.sol
+++ b/contracts/contracts/Sapphire.sol
@@ -237,11 +237,8 @@ library Sapphire {
      *
      * Will cause a reversion if the current usage is more than the amount
      */
-    function gaspad(uint128 toAmount) internal view
-    {
-        (bool success,) = GAS_PAD.staticcall(
-            abi.encode(toAmount)
-        );
+    function gaspad(uint128 toAmount) internal view {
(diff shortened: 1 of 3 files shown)


matevz:
I'm considering the threat model of the gas attacks. By using the new `gaspad()`, an external observer really cannot know how much gas the actual transaction spent. But, by calling `gaspad()` on an adversarial node, the attacker could see that the call was being made (the attacker knows the location of the gaspad() function inside the sapphire runtime), but he cannot determine *how much* gas was padded. Am I correct?

CedarMist:
I'm not sure about the threat model for an adversarial node.

contracts/contracts/Sapphire.sol
```diff
@@ -23,6 +23,10 @@ library Sapphire {
         0x0100000000000000000000000000000000000007;
     address internal constant CURVE25519_PUBLIC_KEY =
         0x0100000000000000000000000000000000000008;
+    address internal constant GAS_USED =
+        0x0100000000000000000000000000000000000009;
+    address internal constant GAS_PAD =
```
matevz, line 28: This should probably be consistent with https://github.com/oasisprotocol/oasis-sdk/blob/main/runtime-sdk/modules/evm/src/precompile/gas.rs#L23

```suggestion
    address internal constant PAD_GAS =
```

contracts/contracts/Sapphire.sol
```diff
@@ -221,6 +225,28 @@ library Sapphire {
...
     }
+
+    /**
+     * @dev Set the current transactions gas usage to a specific amount
+     * @param toAmount Gas usage will be set to this amount
+     * @custom:see @oasisprotocol/oasis-sdk :: precompile/gas.rs :: call_pad_gas
+     *
+     * Will cause a reversion if the current usage is more than the amount
+     */
+    function gaspad(uint128 toAmount) internal view {
```
matevz, line 236:
```suggestion
    function padGas(uint128 toAmount) internal view {
```

contracts/contracts/Sapphire.sol
```diff
@@ -221,6 +225,28 @@ library Sapphire {
...
+        (bool success, ) = GAS_PAD.staticcall(abi.encode(toAmount));
+        require(success, "verify: failed");
+    }
+
+    /**
+     * @dev Returns the amount of gas currently used by the transaction
+     * @custom:see @oasisprotocol/oasis-sdk :: precompile/gas.rs :: call_gas_used
+     */
+    function gasused() internal view returns (uint64) {
```
matevz, line 245:
```suggestion
    function gasUsed() internal view returns (uint64) {
```

matevz (changes requested):
Can you make the naming consistent with the rust code?

matevz (approved):
We should also add this to:
- https://github.com/oasisprotocol/docs/blob/main/docs/dapp/sapphire/precompiles.md
- extend our docs chapter to cover the usage of gas padding (https://docs.oasis.io/dapp/sapphire/guide#writing-secure-dapps ? )
- extend one of the examples to also cover this (demo-voting, demo-starter?) or write a new one

contracts/test/gas.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    let tx = await contract.testConstantTime(1, 100000);
+    let receipt = await tx.wait();
+    expect(receipt.cumulativeGasUsed).eq(expectedGas);
+
+    tx = await contract.testConstantTime(2, 100000);
+    receipt = await tx.wait();
+    expect(receipt.cumulativeGasUsed).eq(expectedGas);
+
+    tx = await contract.testConstantTime(1, 100000);
```
aefhm, line 29: This asserts that we are not caching `receipt.cumulativeGasUsed` right?

contracts/test/gas.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('Gas Padding works as Expected', async () => {
+    const expectedGas = 122735;
+
+    let tx = await contract.testConstantTime(1, 100000);
+    let receipt = await tx.wait();
+    expect(receipt.cumulativeGasUsed).eq(expectedGas);
+
+    tx = await contract.testConstantTime(2, 100000);
```
aefhm, line 25: `contract.testConstantTime(2, 100000);` asserts non-hashing costs right?

contracts/test/gas.ts
CedarMist: It's a transaction receipt, so the receipt details will be unique to that specific tx.

contracts/test/gas.ts
CedarMist: The tests verify the behavior with both zero and non-zero values being included in the gas cost.

PR #217, opened 2023-10-26
contracts: add SHA384 hash and P-384 curve

CedarMist:
Have tested this against sapphire-dev from https://github.com/oasisprotocol/oasis-web3-gateway/pull/469

### DIFF
--- a/contracts/contracts/Sapphire.sol
+++ b/contracts/contracts/Sapphire.sol
@@ -267,7 +267,7 @@ function sha512(bytes memory input) view returns (bytes memory output) {
  * Hash the input data with SHA-384
  *
  * @custom:standard https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf
- * @custom:see @oasisprotocol/oasis-sdk :: precompile/sha512.rs :: call_sha512
+ * @custom:see @oasisprotocol/oasis-sdk :: precompile/sha2.rs :: call_sha384
  * @param input Bytes to hash
  * @return output 48 byte digest
  */
--- a/contracts/contracts/Sapphire.sol
+++ b/contracts/contracts/Sapphire.sol
@@ -263,7 +263,6 @@ function sha512(bytes memory input) view returns (bytes memory output) {
     require(success, "sha512");
 }
 
-
 /**
  * Hash the input data with SHA-384
  *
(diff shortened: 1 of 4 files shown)


contracts/contracts/Sapphire.sol
```diff
@@ -52,7 +54,9 @@ library Sapphire {
         // Sr25519 signature over the provided message.
         Sr25519,
         // Secp256r1 signature over the provided SHA-256 digest.
-        Secp256r1PrehashedSha256
+        Secp256r1PrehashedSha256,
+        // Secp384r1 signature over the provided SHA-384 digest.
```
matevz, line 58: I'm not sure why this description and of other SigningAlg values is not shown at https://api.docs.oasis.io/sol/sapphire-contracts/contracts/Sapphire.sol/library.Sapphire.html#signingalg

Can you use `@dev` tag instead?

contracts/contracts/Sapphire.sol
```diff
@@ -29,6 +29,8 @@ library Sapphire {
         0x0100000000000000000000000000000000000101;
     address internal constant SHA512 =
         0x0100000000000000000000000000000000000102;
+    address internal constant SHA384 =
```
matevz, line 32: 1. While you're at it, can you document this with `@dev` tag and point to the sapphire-paratime rust code where this is implemented? And the same for other constants ;)
2. This adds another 20-byte value to the contract code. Any considerations about the Sapphire contract size? Not sure, if solidity supports this, but could we use a single instance of the contract deployed by ourselves and then have others just linking to it?

matevz (approved):
Some docs suggestions from my side. Otherwise looks good.

contracts/contracts/Sapphire.sol
CedarMist: Re: 2. as it's a `constant` it won't increase contract size. When `Sapphire.sol` is included the optimizer will only emit bytecode for the functions which are called.

contracts/contracts/Sapphire.sol
CedarMist: Annotating individual enum values with NatSpec comments doesn't seem to make any difference to the docs generated by `forge` :(

However, it is possible to annotate the enum itself, e.g.:
```solidity
/// Docs Go here
enum SigningAlg { ... }
```

matevz:
The new precompile should also be added to https://github.com/oasisprotocol/docs/blob/main/docs/dapp/sapphire/precompiles.md

Issue #63, opened 2022-09-29
Create a wrapper library for web3.py


FishmanL:
Title -- should just be a matter of wrapping the transaction generation side and swapping the API url?

PR #218, opened 2023-10-27
clients/python

CedarMist:
Closes: #63 

Example usage:

```python
from sapphirepy import sapphire
w3 = Web3(Web3.HTTPProvider('http://localhost:8545/'))
w3.middleware_onion.add(construct_sign_and_send_raw_middleware(account))
w3 = sapphire.wrap(w3)
```

The `sapphire.wrap` call must be added *after* any `construct_sign_and_send_raw_middleware` or similar signing middleware.

It uses the `epoch` flag for encrypted envelopes, and handles re-fetching the calldata keypair if the epoch is too old. This isn't optimal behavior as if the client runs for any amount of time it will end up using an old keypair for calldata encryption as it doesn't know when the validity expires without first receiving an error.


.github/workflows/contracts-test.yaml
```diff
@@ -28,6 +28,11 @@ jobs:
         run: curl -kv http://127.0.0.1:8545/ || true
       - name: Checkout code
         uses: actions/checkout@v4
+      - name: Python client tests
```
aefhm, line 31: Should we also lint?

clients/py/sapphirepy/deoxysii.py
```diff
@@ -0,0 +1,580 @@
+"""
+Copyright (c) 2019 Oasis Labs Inc. <info@oasislabs.com>
```
aefhm, line 2: I see the reference [here](https://github.com/oasisprotocol/deoxysii/blob/2091330c22b71dd4e04e15306498e68f3a1b3afc/impl.go#L1), but perhaps this should be OPF?

clients/py/sapphirepy/tests/testdata/Deoxys-II-256-128-official-20190608.json
```diff
@@ -0,0 +1,66 @@
...
+		"Sealed":         "88294fcef65a1bdfd7baaa472816c64ef5bef2622b88c1ec5a739396157ef4935f3aa76449e391c32da28ee2857f399ac3dd95aed30cfb26cc0063cd4cd8f7431108176fbf370123856662b000a8348e5925fbb97c9ec0c737758330a7983f06b51590c1d2f5e5faaf0eb58e34e19e5fc85cec03d3926dd46a79ba7026e83dec24e07484c9103dd0cdb0edb505500caca5e1d5dbc71348cf00648821488ebaab7f9d84bbbf91b3c521dbef30110e7bd94f8dad5ab8e0cc5411ca9682d210d5d80c0c4bdbba8181789a4273d6deb80899fdcd976ca6f3a9770b54305f586a04256cfbeb4c11254e88559f294db3b9a94b80ab9f9a02cb4c0748de0af7818685521691dba5738be546dba13a56016fb8635af9dff50f25d1b17ad21707db2640a76a741e65e559b2afaaec0f37e18436bf02008f84dbd7b2698687a22376b65dc7524fca8a28709eee3f3caee3b28ed1173d1e08ee849e2ca63d2c90d555755c8fbafd5d2f4b37f06a1dbd6852ee2ffcfe79d510152e98fc4f3094f740a4aede9ee378b606d34576776bf5f1269f5385a84b3928433bfca177550ccfcd22cd0331bbc595e38c2758b2662476fa66354c4e84c7b360405aa3f5b2a48621bdca1a90c69b21789c91b5b8c568e3c741d99e22f6d7e26f2abed045f1d578b782ab4a5cf2af636d842b3012e180e4b045d8d15b057b69c92398a517053daf9be7c2935ea616f0c218e18b526cf2a3f8c115e262"
+	}
+]
```
aefhm, line 66:
```suggestion
]

```

clients/py/sapphirepy/tests/testdata/Greeter.sol
```diff
@@ -0,0 +1,35 @@
...
+    function revertWithReason() external pure {
+        require(false, "reasonGoesHere");
+    }
+
+    error MyCustomError(string blah);
+
+    function revertWithCustomError() external pure {
+        revert MyCustomError("thisIsCustom");
+    }
+}
```
aefhm, line 35:
```suggestion
}

```

clients/py/sapphirepy/deoxysii.py
```diff
@@ -0,0 +1,580 @@
...
+
+        s0 = t0
+        s1 = t1
+        s2 = t2
+        s3 = t3
+        i += 1
+
+    ciphertext[:16] = struct.pack('>LLLL', s0, s1, s2, s3)
+
+class DeoxysII:
```
aefhm, line 375: I hate to be that guy...but should we repo this separately à la Go [package](https://github.com/oasisprotocol/deoxysii)?

clients/py/sapphirepy/sapphire.py
```diff
@@ -0,0 +1,135 @@
+from typing import Any, Callable, cast, TypedDict
+from binascii import unhexlify, hexlify
+
+from web3 import Web3
+from web3.types import RPCEndpoint, RPCResponse, TxParams
+from eth_typing import HexStr
+
+from .envelope import TransactionCipher
+
+# Should transactions which deploy contracts be encrypted?
```
aefhm, line 10: > "When it comes to privacy and accountability, people always demand the former for themselves and the latter for everyone else." - David Brin

clients/py/sapphirepy/envelope.py
```diff
@@ -0,0 +1,128 @@
...
+        if not isinstance(call_result, dict):
+            raise EnvelopeError('callResult', type(call_result))
+        if call_result.get('failure', None) is not None:
+            raise CallFailure(call_result['failure'])
+        ok = call_result.get('ok', None)
+        if not isinstance(ok, dict):
+            raise EnvelopeError('callResult.ok', type(ok))
+        if ok is not None:
+            return self._decrypt_inner(ok)
+        raise EnvelopeError("No 'ok in call result!")
```
aefhm, line 128:
```suggestion
        raise EnvelopeError("No 'ok' in call result!")
```

clients/py/sapphirepy/deoxysii.py
CedarMist: For now, it is easiest to just include it in the package as it's a single python file that's only used in one place.

We can consider moving it to an external package in the near future

clients/py/sapphirepy/sapphire.py
CedarMist: By default transactions which deploy code are unencrypted, but there is a manual toggle to turn on deployment transactions.

.github/workflows/contracts-test.yaml
CedarMist: I've added a lint step to CI

clients/py/sapphirepy/deoxysii.py
CedarMist: Eh, it's a direct port of the Go client, so technically a derivative work *shrugs*

clients/py/sapphirepy/sapphire.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self):
+        self._keys = []
+
+    def _trim_and_sort(self, newest_epoch:int):
+        self._keys = sorted([v for v in self._keys
+                             if v['epoch'] >= newest_epoch - EPOCH_LIMIT],
+                            key=lambda o: o['epoch'])[-EPOCH_LIMIT:]
+
+    @property
+    def newest(self):
```
matevz, line 33: I think you can safely use Python >=3.10 and annotate composite return type hints as well.

```suggestion
    def newest(self) -> CalldataPublicKey | None:
```

matevz (commented):
Nice work. I'll give it a spin tomorrow.

matevz:
Is there a library to generate python bindings for your solidity contract (similar to ethers for javascript) so you could do something like:

```py
greeter = Greeter()
greeter.deploy().wait()
print(greeter.address)
```

Then, we could actually make a practical example and add it to our official documentation.

matevz:
> Is there a library to generate python bindings for your solidity contract (similar to ethers for javascript) so you could do something like:
> 
> ```python
> greeter = Greeter()
> greeter.deploy().wait()
> print(greeter.address)
> ```
> 
> Then, we could actually make a practical example and add it to our official documentation.

Ah, it's ethers for python duh https://pypi.org/project/ethers/

CedarMist:
web3.py allows a similar pattern: https://github.com/oasisprotocol/sapphire-paratime/pull/218/files#diff-00b5c90060622faf7d475f14961a2a6b11a9d4d8f1d316ab3c3408864bab5635R45

```python
x = self.greeter.functions.blah().transact({'gasPrice': w3.eth.gas_price})
y = w3.eth.wait_for_transaction_receipt(x)
z = greeter.events.Greeting().process_receipt(y)
```

To support ethers-py we'd either have to add support for `ethers-rs` or add a wrapper for `ethers-py`, most people in Python land seem to use web3.py and `ethers-py` doesn't have much usage.

clients/py/sapphirepy/sapphire.py
CedarMist: I'm already using composite types, and VS Code can determine the type of `newest` so the return type wasn't annotated:

```python
newest: (self: Self@CalldataPublicKeyManager) -> (CalldataPublicKey | None)
```

I should specify 3.10 explicitly in setup.py description though, as it won't run on earlier versions unless I move to `typing.Union` and `typing.Optional` notation

matevz (changes requested):

clients/py/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+# Setup your Web3 provider with a signing account
+w3 = Web3(Web3.HTTPProvider('http://localhost:8545'))
+w3.middleware_onion.add(construct_sign_and_send_raw_middleware(account))
+
+# Finally, wrap the provider to add Sapphire end-to-end encryption
+w3 = sapphire.wrap(w3)
+```
+
+The Sapphire middleware for Web3.py ensures all transactions, gas estimates and view calls are end-to-end encrypted between your application and the smart contract.
```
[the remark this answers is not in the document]
CedarMist: Correct, it does not support signed view queries.

View calls rely on the calldata public key returned from the web3 gateway to prevent tampering. Currently this isn't validated so all clients can be MITMd

clients/py/sapphirepy/sapphire.py
```diff
@@ -0,0 +1,135 @@
...
+            if method == 'eth_call' and result.get('result', '0x') != '0x':
+                decrypted = c.decrypt(unhexlify(result['result'][2:]))
+                result['result'] = HexStr('0x' + hexlify(decrypted).decode('ascii'))
+
+            return result
+        return make_request(method, params)
+    return middleware
+
+def wrap(w3: Web3):
+    if 'sapphire' not in w3.middleware_onion:
```
matevz, line 133: Document this function.

clients/py/setup.py
```diff
@@ -0,0 +1,33 @@
...
+    name="sapphire.py",
+    license="Apache Software License 2.0",
+    long_description=README,
+    long_description_content_type="text/markdown",
+    include_package_data=True,
+    python_requires='>=3.8',
+    packages=find_packages(include=["sapphirepy"]),
+    install_requires=REQUIREMENTS,
+    url="https://github.com/oasisprotocol/sapphire-paratime",
+    version="0.3.0",
```
matevz, line 31: I suppose you want to have 0.3.0 so it's upgradable from the Ocean's 0.2.1?

matevz (approved):
I think this is good to go. We may add some API docs for the python client to api.docs.oasis.io as well in the future.

PR #227, opened 2023-11-24
docs: Migrate Sapphire documentation from oasisprotocol/docs

matevz:
This PR migrates topic-oriented Sapphire documentation from https://github.com/oasisprotocol/docs/tree/main/docs/dapp/sapphire to this repository. The README file is still kept inside the docs repository to be consistent with Emerald and Cipher.


CedarMist (approved):
Ok this looks good, all the docs here are self-contained

matevz:
> Ok this looks good, all the docs here are self-contained

Actually there are 3 images inside guide.mdx which caused me some gray hair and didn't find a way arond it. They are not shown, if not viewed from the docs site. The remaining material works fine. Links to markdown files from other repos (inlcuding the docs repo) must be provided as https://github.com/oasisprotocol/...) and the documentation parser on docs.oasis.io will transform it to a local path.

PR #231, opened 2023-11-28
Add Security documentation to ensure finality

aefhm:
#### Description

Relates to https://github.com/oasisprotocol/docs/issues/509#issuecomment-1799192990

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -76,6 +76,36 @@ function transferFrom(address who, address to, uint amount)
 }
 ```
 
+## Speed Bump
+
+If we would like to prevent off-chain calls from being chained together, we can
+ensure that the block has been finalized.
+
+```solidity
+contract Secret {
+  uint256 private _height;
+  bytes private _secret;
+  address private _buyer;
+
+  constructor(bytes memory _text) {
+    _secret = _text;
+  }
+
+  function recordPayment() external payable {
+    require(msg.value == 1 ether);
+    // set and lock recipient
+    _height = block.number;
+  }
+
+  /// @notice Reveals the secret.
+  function revealSecret() view external returns (bytes memory) {
+    require(block.number >= _height, "not settled");
+    // check for recipient
+    return _secret;
+  }
+}
+```
+
 ## Gas Padding
 
 To prevent leaking information about a particular transaction, Sapphire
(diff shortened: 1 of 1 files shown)


kostko:
Task linked: [CU-86934gu0x Add security notice for block checking](https://app.clickup.com/t/86934gu0x)

docs/security.md
```diff
@@ -76,6 +76,36 @@ function transferFrom(address who, address to, uint amount)
...
+ensure that the block has been finalized.
+
+```solidity
+contract Secret {
+  uint256 private _height;
+  bytes private _secret;
+  address private _buyer;
+
+  constructor(bytes memory _text) {
+    _secret = _text;
```
kostko, line 91: Should also initialize `_height` to some maximum value or have an additional flag (maybe `_buyer` being non-equal to `address(0)`) as otherwise `_height` is initialized to zero and every block number will be greater than that.

docs/security.md
```diff
@@ -76,6 +76,36 @@ function transferFrom(address who, address to, uint amount)
...
+## Speed Bump
+
+If we would like to prevent off-chain calls from being chained together, we can
+ensure that the block has been finalized.
+
+```solidity
+contract Secret {
+  uint256 private _height;
+  bytes private _secret;
+  address private _buyer;
```
kostko, line 88: `_buyer` seems unused, so it can just be removed (or used as suggested).

docs/security.md
aefhm: Makes sense. Opting to check `_buyer`.

docs/security.md
```diff
@@ -76,6 +76,37 @@ function transferFrom(address who, address to, uint amount)
...
+    require(msg.value == 1 ether);
+    // set and lock buyer
+    _height = block.number;
+    _buyer = msg.sender;
+  }
+
+  /// @notice Reveals the secret.
+  function revealSecret() view external returns (bytes memory) {
+    require(block.number > _height, "not settled");
+    require(_buyer != address(0), "no recorded buyer");
```
kostko, line 104: Sorry for more nits, but you could add a comment now saying something like `// TODO: Optionally authenticate buyer's call.` as currently the secret would be revealed to anyone (which is fine to keep this example small, we should just make sure to note that the example is incomplete).

aefhm: Nits welcome! 🫶

Yeah, I think the warning is nice.

kostko (approved):

PR #236, opened 2023-12-01
ci: fix NPM publish + minor contracts tests fix

CedarMist:
This requires PNPM to install dependencies first
Bumps clients/js version to v1.2.2
Fixes gas test in contracts (sometimes it's... 1 gas less?)

### DIFF
--- a/clients/js/package.json
+++ b/clients/js/package.json
@@ -2,7 +2,7 @@
   "type": "module",
   "name": "@oasisprotocol/sapphire-paratime",
   "license": "Apache-2.0",
-  "version": "1.2.0",
+  "version": "1.2.2",
   "description": "The Sapphire ParaTime Web3 integration library.",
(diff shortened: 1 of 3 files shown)


matevz (approved):

clients/js/package.json
```diff
@@ -2,7 +2,7 @@
   "type": "module",
   "name": "@oasisprotocol/sapphire-paratime",
   "license": "Apache-2.0",
-  "version": "1.2.0",
+  "version": "1.2.2",
```
aefhm, line 5: Not `1.2.1`?

aefhm (approved):

clients/js/package.json
CedarMist: I attempted to release 1.2.1 via a tag but it failed.
Unable to remove tags or force push tags, so using 1.2.2

PR #261, opened 2024-01-23
Fixes some minor issues to make it work with Ethers v5 & v6, 

CedarMist:
and reduces bundle size by adding ethers as a peer dependency

js-sha512 is replaced with `@noble/hashes` (used by Ethers v6)

Both ethers & noble hashes are added as peer dependencies.

Some minor fixes around detecting the chain ID and whether it's ethers v6. If direct type comparisons are done (e.g. `instanceof AbstractSigner` and the ethers version isn't identical it will fail, so ended up using reflection to roughly determine if it's ethers v5 or ethers v6)


clients/js/src/compat.ts
```diff
@@ -590,61 +594,83 @@ export async function fetchRuntimePublicKeyByChainId(
...
-        };
         resp = await source.send(OASIS_CALL_DATA_PUBLIC_KEY, []);
       } catch (ex) {
         // don't do anything, move on to try chainId fetch
+        chainId = fromQuantity(await source.send('eth_chainId', []));
       }
     }
+    // Otherwise, we have no idea what to do with this provider!
+    else {
+      throw new Error(
```
aefhm, line 662: We don't want to default to fetch by `chainId` here?

clients/js/src/compat.ts
```diff
@@ -590,61 +594,83 @@ export async function fetchRuntimePublicKeyByChainId(
...
     }
   }
 
-  // Note: MetaMask does not support Web3 methods it doesn't know about, so we have to
-  // fall back to manually querying the default gateway.
-  const chainId = Number(
-    (await new BrowserProvider(provider).getNetwork()).chainId,
-  );
+  if (!chainId) {
+    throw new Error('fetchRuntimePublicKey to retrieve chainId from provider');
```
aefhm, line 673:
```suggestion
    throw new Error('fetchRuntimePublicKey failed to retrieve chainId from provider');
```

clients/js/src/compat.ts
```diff
@@ -590,61 +594,83 @@ export async function fetchRuntimePublicKeyByChainId(
...
         resp = await source.send(OASIS_CALL_DATA_PUBLIC_KEY, []);
       } catch (ex) {
         // don't do anything, move on to try chainId fetch
+        chainId = fromQuantity(await source.send('eth_chainId', []));
       }
     }
+    // Otherwise, we have no idea what to do with this provider!
+    else {
+      throw new Error(
+        'fetchRuntimePublicKey not sure how to handle non-request non-send provider!',
```
aefhm, line 663: Dunno if we want to be more firm 🤔 

```suggestion
        'fetchRuntimePublicKey does not support non-request non-send provider!',
```

clients/js/package.json
```diff
@@ -65,5 +65,9 @@
     "ts-node": "^10.9.2",
     "typedoc": "^0.25.4",
     "typescript": "^4.8.3"
+  },
+  "peerDependencies": {
+    "ethers": "6.x",
```
aefhm, line 70: Hmm. Interesting that this works with v5.

clients/js/package.json
```diff
@@ -2,7 +2,7 @@
   "type": "module",
   "name": "@oasisprotocol/sapphire-paratime",
   "license": "Apache-2.0",
-  "version": "1.2.2",
+  "version": "1.2.3",
```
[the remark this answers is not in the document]
CedarMist: Bump to 1.3

clients/js/package.json
CedarMist: The idea is that if ethers v6 is installed it will use the same version.

I guess alternatively we could change the ethers dependency to `6.x` and remove the peerDependency.

If the users app uses ethers v5 and they install sapphire-paratime package their bundled dApp will still end up bundling both v5 & v6 which seems unavoidable at this point unless we completely remove ethers from sapphire-paratime...

This should be possible if we remove:

```
if (typeof upstream === 'string') {
    upstream = new JsonRpcProvider(upstream) as any;
  }
```

And then extract the needed types from ethers, and the utility functions `hexlify`, `isBytesLike`, `getBytes` etc.

CedarMist: I did some reading and I'm still unsure of best practice for ensuring that if ethers v6 is installed we use whatever version the dapp is using

clients/js/src/compat.ts
```diff
@@ -122,6 +122,7 @@ export function wrap<U extends UpstreamProvider>(
     return wrapEIP1193Provider(upstream, filled_options);
   }
 
+  console.log(upstream);
```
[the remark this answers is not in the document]
CedarMist: Arghh facepalm...

lubej (approved):

clients/js/package.json
aefhm: Agree to stay with status quo for now.

clients/js/src/compat.ts
```diff
@@ -292,8 +292,11 @@ function isEthers5Signer(upstream: object): upstream is Ethers5Signer {
 }
 
 function isEthers6Signer(upstream: object): upstream is Signer {
-  // XXX: this will not match if installed ethers version is different!
-  return upstream instanceof AbstractSigner;
+  return (
+    upstream instanceof AbstractSigner ||
+    (Reflect.get(upstream, 'signTypedData') &&
+      Reflect.get(upstream, 'signTransaction'))
```
lukaw3d, line 298: How come you're using Reflect.get, not `upstream.signTypedData && upstream.signTransaction`?

CedarMist: There is no specific reason.

lukaw3d: Reflect.get basically ignores type safety and returns `any`

PR #262, opened 2024-01-23
chore: Refactor code

lubej:
- add v5 ethers types;
- improve typings;


CedarMist:
This adds Ethers v5 back in after it's been carefully removed, so the package goes back to relying on both ethers v5 & ethers v6.

lubej:
Closing in favor of https://github.com/oasisprotocol/sapphire-paratime/pull/261

PR #270, opened 2024-02-05
Refactor callDataPublicKey handling

CedarMist:
This should fix several long-standing issues with call encryption.

Previously the `Cipher` object would be created and maintained with the same keys throughout the duration of the session, it also did not support the new `Epoch` flag which specifies which calldata public key was being used to encrypt the request (requiring the server to try several).

This would cause problems, like when you leave a dApp open for a while your transactions would fail with 'tag verification failed'.

This introduces several changes:

 * New Cipher instance is used for every request
 * calldata public key is cached (5 minutes by default)
 * The `epoch` parameter is passed in the encryption envelope
 * Debug `scripts/proxy.js` shows epoch status too.

This means that leaving dApps open for prolonged periods of time will be OK.

---------------------

Follow-up notes:

 * Gas tests in `contracts/` had a small rewrite, because of the variance they were longer a sufficient test of what they were supposed to do check for! They've been made relative instead.

### DIFF
--- a/clients/js/src/interfaces.ts
+++ b/clients/js/src/interfaces.ts
@@ -77,3 +77,8 @@ export async function undefer<T>(obj: Deferrable<T>): Promise<T> {
     await Promise.all(Object.entries(obj).map(async ([k, v]) => [k, await v])),
   );
 }
+
+export type UpstreamProvider =
+  | EIP1193Provider
+  | Ethers5Signer
+  | Ethers5Provider;
--- a/clients/js/package.json
+++ b/clients/js/package.json
@@ -2,7 +2,7 @@
   "type": "module",
   "name": "@oasisprotocol/sapphire-paratime",
   "license": "Apache-2.0",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "description": "The Sapphire ParaTime Web3 integration library.",
   "homepage": "https://github.com/oasisprotocol/sapphire-paratime/tree/main/clients/js",
   "repository": {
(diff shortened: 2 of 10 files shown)


clients/js/src/interfaces.ts
```diff
@@ -77,3 +77,8 @@ export async function undefer<T>(obj: Deferrable<T>): Promise<T> {
     await Promise.all(Object.entries(obj).map(async ([k, v]) => [k, await v])),
   );
 }
+
+export type UpstreamProvider =
+  | EIP1193Provider
```
lubej, line 82: Think this needs EIP1193 Signer as well.

CedarMist: The EIP-1193 interface is just the interface used to communicate with the Web3 wallet, make JSON-RPC requests, event notification etc. - so all operations are accessed through it, rather than having separate Provider/Signer interfaces like with Ethers.

See: https://eips.ethereum.org/EIPS/eip-1193

lubej: What confused me was `Ethers5Signer` listed as one of the options of `UpstreamProvider`.

CedarMist: The Ethers5Signer / Ethers5Provider types were extracted from the ethers package so as to not require importing both ethers v5 and ethers v6. For Ethers it needs to interact with them an entirely different and more complex way than with EIP-1193 compatible providers.

lubej (approved):
LGTM

PR #276, opened 2024-02-19
examples: added require() reversion message check to ethers v6 client

CedarMist:
Double check that reversion with error message works with Ethers v6

### DIFF
--- a/examples/ethersv6-ts-esm/src/index.ts
+++ b/examples/ethersv6-ts-esm/src/index.ts
@@ -16,6 +16,8 @@ async function testTheContract(contract:BaseContract, signerAddr?:string)
         assert( addr === ZeroAddress );
     }
 
+    // Verifies that calling a function which does require(false,"ThisIsAnError")
+    // Will return the correct error message in the exception to Ethers
     try {
         await contract.getFunction("testViewRevert()").staticCall();
         assert(false);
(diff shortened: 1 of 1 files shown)


examples/ethersv6-ts-esm/src/index.ts
```diff
@@ -16,6 +16,15 @@ async function testTheContract(contract:BaseContract, signerAddr?:string)
         assert( addr === ZeroAddress );
     }
 
+    try {
+        await contract.getFunction("testViewRevert()").staticCall();
```
matevz, line 20: Can you put a comment on what should this suppose to test? Querying unexisting methods should raise error?

matevz (approved):

examples/ethersv6-ts-esm/src/index.ts
aefhm: ➕ to adding more context. I think this method does exist and is specifically there to error out?

CedarMist: I've added some comments.

Yes, it's to test that errors which are returned by smart contracts are correctly handled by the Ethers v6 wrapped client library. I'm testing both `require(false, "message")` and `revert` with a custom error type.

This is important for DX, and highlights that Ethers v5 handles errors in a weird way (meaning you can't tell the difference between data returned from a view call, or data exposed in a revert...) which was subsequently fixed in Ethers v6.

An example of an Ethers v5 bug / exploit:

 If I had a smart contract, lets call it 'canTransferAllTheMoney' which calls another smart contract, in the other smart contract I could revert in a way which returns data that looks like the function returns `true` when in fact a contract it called had raised an error.

Whereas with Ethers v6 this will correctly be interpreted as an error, so there's no chance of confusion in the UI.

PR #278, opened 2024-02-21
clients/go: Bump oasis-core to 23.0

matevz:
This PR:
- Bumps clients/go to use oasis-core 23.x types.
- removes redundant data types which are already part of the oasis-sdk/client-sdk/go
- removes SendTransaction reimplementation
- fixes signed query implementation
- implements encrypted and signed eth_estimateGas calls
- adds test for PackSignedCall()


clients/go/compat.go
```diff
@@ -6,12 +6,12 @@ import (
...
 	"sync"
 
-	"github.com/ethereum/go-ethereum"
+	ethereum "github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
-	"github.com/fxamacker/cbor/v2"
+	cbor "github.com/fxamacker/cbor/v2"
```
kostko, line 14: This should not be needed, also you should use CBOR from Oasis Core to make sure the correct version and settings are used.

clients/go/go.mod
```diff
@@ -1,8 +1,13 @@
 module github.com/oasisprotocol/sapphire-paratime/clients/go
 
-go 1.19
+go 1.21
```
aefhm, line 3: (nit) workflow is 1.22?

clients/go/cipher.go
```diff
@@ -30,30 +32,8 @@ var (
...
-}
-
-type AeadEnvelope struct {
-	Nonce []byte `json:"nonce"`
-	Data  []byte `json:"data"`
-}
-
 type Cipher interface {
-	Kind() uint64
+	Kind() types.CallFormat
```
peternose, line 36:
```suggestion
	CallFormat() types.CallFormat
```

clients/go/cipher.go
```diff
@@ -67,34 +47,34 @@ func NewPlainCipher() PlainCipher {
 	return PlainCipher{}
 }
 
-func (c PlainCipher) Kind() uint64 {
-	return Plain
+func (c PlainCipher) Kind() types.CallFormat {
+	return types.CallFormatPlain
```
peternose, line 51: It feels like call format is cipher specific. So maybe this change is not the best, as if I want to add another cipher, I need to update the client-sdk library first.

clients/go/cipher.go
```diff
@@ -67,34 +47,34 @@ func NewPlainCipher() PlainCipher {
...
 
-func (c PlainCipher) Decrypt(nonce []byte, ciphertext []byte) (plaintext []byte, err error) {
+func (c PlainCipher) Decrypt(_ []byte, ciphertext []byte) (plaintext []byte, err error) {
 	return ciphertext, nil
 }
 
 func (c PlainCipher) DecryptCallResult(response []byte) ([]byte, error) {
-	var callResult CallResult
+	var callResult types.CallResult
 	cbor.MustUnmarshal(response, &callResult)
```
peternose, line 65: This still needs to be addressed, as discussed offline.

clients/go/cipher.go
```diff
@@ -205,54 +186,58 @@ func (c X25519DeoxysIICipher) EncryptEncode(plaintext []byte) []byte {
...
-	if callResult.Fail != nil {
+	if callResult.Failed != nil {
 		return nil, ErrCallFailed
 	}
 
-	var aeadEnvelope AeadEnvelope
-	if callResult.OK != nil {
-		if err := cbor.Unmarshal(callResult.OK, &aeadEnvelope); err != nil {
-			return callResult.OK, nil
+	var aeadEnvelope types.ResultEnvelopeX25519DeoxysII
```
peternose, line 199: Same here. I would probably expect that cipher defines an envelope, not an external library.

clients/go/cipher.go
kostko: But to add a new cipher you need to update the SDK anyway?

clients/go/compat.go
```diff
@@ -57,6 +63,7 @@ func PackTx(tx types.Transaction, cipher Cipher) (*types.Transaction, error) {
 }
 
 func packTx(tx types.Transaction, cipher Cipher) (*types.Transaction, error) {
+	fmt.Printf("NONCE: %d\n", tx.Nonce())
```
aefhm, line 66: (nit maybe) Do we always want to log?

kostko: This should be removed, likely a debugging leftover :)

clients/go/cipher.go
kostko: Yes, agreed, both should be in the SDK.

clients/go/compat.go
kostko, line 1: Unrelated to this PR, but why is `WrapClient` not taking an `*ethclient.Client` so one needs to make a shallow copy and then `WrapClient` takes its address internally anyway?

Also it seems like `WrappedBackend` needs to implement a bunch of interfaces, but there are no asserts to make sure.

Also what's with the global `packedTxs` map which seems to be an ever-growing cache of transaction hashes?

kostko (commented):
This thing still needs some refactoring.

kostko:
This current API seems confusing and error-prone. It looks like it has two hooks to manipulate transactions:
* It can generate a `TransactOpts` instance with a wrapped signer that also performs encryption.
* It then also provides an wrapped `SendTransaction` which performs mangling, but in this case it is already too late in the pipeline to actually return the modified transaction! So in case this second fallback is used the transaction that has actually been sent to Sapphire cannot be retrieved by the developer. Then there is additional hacky logic to detect whether a transaction was previously mangled by the wrapped signer so that it is not wrapped again.

I think it would be better if `SendTransaction` was not overriden and in case you want encrypted transaction you either need to encrypt them manually or use the `Transactor`.

clients/go/README.md
```diff
@@ -1,63 +1,79 @@
...
+signedTx := sign(packedTx) // Using your usual signer.
 ```
 
-and sending it with a normal, not-wrapped `ethclient`:
+and sending it with a normal, not-wrapped `ethclient.Client` instance:
 
-```Go
-ethclient.SendTransaction(ctx, signedTx)
+```go
+_ := c.SendTransaction(ctx, signedTx)
```
kostko, line 76:
```suggestion
_ = c.SendTransaction(ctx, signedTx)
```

clients/go/compat.go
```diff
@@ -72,13 +83,34 @@ func PackCall(msg ethereum.CallMsg, cipher Cipher) (*ethereum.CallMsg, error) {
...
+
+func (s rsvSigner) SignRSV(digest [32]byte) ([]byte, error) {
+	return s.sign(digest)
+}
+
 // PackSignedCall prepares `msg` in-place for being sent to Sapphire. The call will be end-to-end encrypted and a signature will be used to authenticate the `from` address.
-func PackSignedCall(msg ethereum.CallMsg, cipher Cipher, sign SignerFn, chainID big.Int, leash Leash) (*ethereum.CallMsg, error) {
-	dataPack, err := NewDataPack(sign, chainID.Uint64(), msg.From[:], msg.To[:], 0, msg.GasPrice, msg.Value, msg.Data, leash)
+func PackSignedCall(msg ethereum.CallMsg, cipher Cipher, sign SignerFn, chainID big.Int, leash evm.Leash) (*ethereum.CallMsg, error) {
+	fmt.Printf("PackSignedCall data: %x\n", msg.Data)
```
kostko, line 96: Remove.

clients/go/compat.go
```diff
@@ -100,18 +132,19 @@ func NewWrappedBackend(backend bind.ContractBackend, chainID big.Int, cipher Cip
...
 	keypair, err := NewCurve25519KeyPair()
 	if err != nil {
 		return nil, fmt.Errorf("failed to generate ephemeral keypair: %w", err)
 	}
-	cipher, err := NewX25519DeoxysIICipher(*keypair, *runtimePublicKey)
+	cipher, err := NewX25519DeoxysIICipher(keypair, runtimePublicKey, epoch)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create default cipher: %w", err)
 	}
+	//cipher := NewPlainCipher()
```
kostko, line 147: Remove.

clients/go/compat.go
```diff
@@ -175,25 +208,34 @@ func (b WrappedBackend) CallContract(ctx context.Context, call ethereum.CallMsg,
...
+			BlockRange:  DefaultBlockRange,
+		}
+		call.Gas = math.MaxInt64        // Must be non-zero for signing to work.
+		call.GasPrice = big.NewInt(0x1) // Must be non-zero for signing to work.
 		if packedCall, err = PackSignedCall(call, b.cipher, b.sign, b.chainID, leash); err != nil {
 			return nil, fmt.Errorf("failed to pack signed call: %w", err)
 		}
 	}
+	pc, _ := json.Marshal(packedCall)
+	fmt.Printf("packed call: %s\n", pc)
```
kostko, line 238: Remove.

clients/go/compat.go
```diff
@@ -175,25 +208,34 @@ func (b WrappedBackend) CallContract(ctx context.Context, call ethereum.CallMsg,
...
-		blockHash := header.Hash()
-		leash := NewLeash(header.Nonce.Uint64(), header.Number.Uint64(), blockHash[:], DefaultBlockRange)
+		// We will build a leash on the pre-last block.
+		blockHash := header.ParentHash
+		leashBlockNumber.Sub(header.Number, big.NewInt(1))
+		nonce, err := b.backend.PendingNonceAt(ctx, call.From)
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch account nonce: %w", err)
+		}
+		fmt.Printf("BLOCK HASH: %x\n", blockHash)
```
kostko, line 222: Remove.

clients/go/compat.go
```diff
@@ -175,25 +208,34 @@ func (b WrappedBackend) CallContract(ctx context.Context, call ethereum.CallMsg,
...
+		headerJson, _ := json.Marshal(header)
+		fmt.Printf("BLOCK HEADER: %s\n", headerJson)
```
kostko, line 224: Remove.

clients/go/compat.go
```diff
@@ -175,25 +208,34 @@ func (b WrappedBackend) CallContract(ctx context.Context, call ethereum.CallMsg,
...
+		leash := evm.Leash{
+			Nonce:       nonce,
+			BlockNumber: leashBlockNumber.Uint64(),
+			BlockHash:   blockHash[:],
+			BlockRange:  DefaultBlockRange,
+		}
+		call.Gas = math.MaxInt64        // Must be non-zero for signing to work.
+		call.GasPrice = big.NewInt(0x1) // Must be non-zero for signing to work.
```
kostko, line 232: Why are these overridden to some magic values? I would only override in case these are not set in `call` and these should be constants.

kostko (commented):
As mentioned above, I think we should remove the hacky `SendTransaction` behavior unless there is a good reason to keep it.

clients/go/README.md
```diff
@@ -1,63 +1,93 @@
...
+
+```go
+import (
+    "context"
+
+    "github.com/ethereum/go-ethereum/accounts/abi/bind"
+    "github.com/ethereum/go-ethereum/ethclient"
+	
+	sapphire "github.com/oasisprotocol/sapphire-paratime/clients/go"
+)
```
kostko, line 34: Inconsistent indentation.

clients/go/compat.go
```diff
@@ -228,40 +249,60 @@ func (b WrappedBackend) SuggestGasTipCap(ctx context.Context) (*big.Int, error)
...
+	return &evm.Leash{
+		Nonce:       nonce,
+		BlockNumber: leashBlockNumber.Uint64(),
+		BlockHash:   blockHash[:],
+		BlockRange:  DefaultBlockRange,
+	}, nil
 }
 
 func txNeedsPacking(tx *types.Transaction) bool {
 	_, isPacked := packedTxs.Load(tx.Hash())
```
kostko, line 295: Why do we need `packedTxs`? Can't we infer that from the transaction itself? This seems hacky and also a potential memory leak with `packedTxs` not getting cleaned up.

clients/go/compat.go
matevz: > Unrelated to this PR, but why is `WrapClient` not taking an `*ethclient.Client` so one needs to make a shallow copy and then `WrapClient` takes its address internally anyway?

Fixed.

> Also it seems like `WrappedBackend` needs to implement a bunch of interfaces, but there are no asserts to make sure.

Yeah, `WrappedBackend` implements `bind.ContractBackend` interface. I changed the return value of `WrapClient()` to make sure.

> Also what's with the global `packedTxs` map which seems to be an ever-growing cache of transaction hashes?

Removed.

clients/go/README.md
```diff
@@ -1,63 +1,93 @@
...
-
-If you have more than a little trouble, please file an issue. There should be
-_no_ reason _not_ to use the Sapphire ParaTime!
+[Sapphire ParaTime]: https://docs.oasis.io/dapp/sapphire
 
 ## Building
 
-Sapphire Paratime compatibility library works with `Go` version 1.17 or later and the latest comparable `go-ethereum` version.
+Sapphire compatibility library works with Go version 1.22 or later and the
+latest comparable `go-ethereum` version.
```
kostko, line 14:
```suggestion
latest compatible `go-ethereum` version.
```

clients/go/README.md
```diff
@@ -1,63 +1,93 @@
...
+instead of the usual `ethclient.Client` instance:
 
-```Go
-txOpts := backend.Transactor(senderAddr)
+```go
 nft, _ := NewNft(addr, backend)
+```
+
+Confidential transactions using Go-Ethereum ABI wrapper can be submitted by
+passing the Sapphire-specific `bind.TransactOpts` as the first parameter:
```
kostko, line 61: Maybe add a warning that in case this transactor is not passed, transactions will be plain-text?

clients/go/compat.go
```diff
@@ -56,6 +62,7 @@ func PackTx(tx types.Transaction, cipher Cipher) (*types.Transaction, error) {
 }
 
 func packTx(tx types.Transaction, cipher Cipher) (*types.Transaction, error) {
+	fmt.Printf("NONCE: %d\n", tx.Nonce())
```
kostko, line 65: Remove this.

clients/go/compat.go
```diff
@@ -100,23 +131,23 @@ func NewWrappedBackend(backend bind.ContractBackend, chainID big.Int, cipher Cip
 
 // NewCipher creates a default cipher.
```
kostko, line 132: Maybe add something that this is only valid for an epoch or similar.

kostko (approved):

aefhm:
I suppose this'll effect a minor version bump?

clients/go/compat.go
```diff
@@ -72,16 +78,40 @@ func PackCall(msg ethereum.CallMsg, cipher Cipher) (*ethereum.CallMsg, error) {
...
+func (s rsvSigner) SignRSV(digest [32]byte) ([]byte, error) {
+	return s.sign(digest)
+}
+
 // PackSignedCall prepares `msg` in-place for being sent to Sapphire. The call will be end-to-end encrypted and a signature will be used to authenticate the `from` address.
-func PackSignedCall(msg ethereum.CallMsg, cipher Cipher, sign SignerFn, chainID big.Int, leash Leash) (*ethereum.CallMsg, error) {
-	dataPack, err := NewDataPack(sign, chainID.Uint64(), msg.From[:], msg.To[:], 0, msg.GasPrice, msg.Value, msg.Data, leash)
+func PackSignedCall(msg ethereum.CallMsg, cipher Cipher, sign SignerFn, chainID big.Int, leash *evm.Leash) (*ethereum.CallMsg, error) {
+	if msg.Gas == 0 {
+		msg.Gas = DefaultGasLimit // Must be non-zero for signed calls.
```
aefhm, line 92: So even `1` will work right? Just confirming as we don't use `DefaultGasLimit` otherwise.

aefhm: 👍 I thought this was overriding the the `msg.Gas`, but it is not.

PR #284, opened 2024-03-13
clients/js: added hook for Ethers JsonRpcApiProvider.getSigner

CedarMist:
When the user calls `.getSigner` on a wrapped Ethers provider, it will return an unwrapped signer.

This change returns a wrapped signer.

### DIFF
--- a/clients/js/src/compat.ts
+++ b/clients/js/src/compat.ts
@@ -247,6 +247,13 @@ export function wrapEthersProvider<P extends Provider | Ethers5Provider>(
     estimateGas: hookEthersCall(provider, 'estimateGas', filled_options),
   };
 
+  if( 'getSigner' in provider ) {
+    hooks['getSigner'] = async function (...args:unknown[]) {
+      const r = await (provider as any).getSigner(...args);
+      return wrapEthersSigner(r);
+    }
+  }
+
   // When a signer is also provided, we can re-pack transactions
   // But only if they've been signed by the same address as the signer
   if (signer) {


matevz (commented):
Wondering if the change is the right thing to do. It seems the current behavior is already incorporated in the demos, for example https://github.com/oasisprotocol/demo-starter/blob/main/frontend/src/stores/ethereum.ts#L97-L121?

Although, I must admit there is too much boilerplate code involved in the demos IMO.

matevz (changes requested):
Rethinking this, you are absolutely right. Can you please add a test to check whether the getSigner() call works as intended?

CedarMist:
> Rethinking this, you are absolutely right. Can you please add a test to check whether the getSigner() call works as intended?

I don't have dappwrite or synpress setup in CI so I can't run E2E tests that require metamask.

Will add a test for `.getSigner()` if it works with Hardhat or with a wallet signed in Ethers, but really need to start testing with metamask otherwise we're missing a lot of coverage.

matevz:
> I don't have dappwrite or synpress setup in CI so I can't run E2E tests that require metamask.
> 
> Will add a test for `.getSigner()` if it works with Hardhat or with a wallet signed in Ethers, but really need to start testing with metamask otherwise we're missing a lot of coverage.

synpress would be ideal, but this can go into another PR. I was just thinking of testing that .getSigner() is defined and that it returns the wrapped instance.

CedarMist:
I literally can't test this without doing it in the browser.

```
Error: could not coalesce error (error={ "code": "UNKNOWN_ERROR", "message": "could not coalesce error (error={ \"code\": -32601, \"payload\": { \"method\": \"eth_requestAccounts\", \"params\": [  ] } }, payload={ \"method\": \"eth_requestAccounts\", \"params\": [  ] }, code=UNKNOWN_ERROR, version=6.10.0)" }, payload={ "id": 2, "jsonrpc": "2.0", "method": "eth_accounts", "params": [  ] }, code=UNKNOWN_ERROR, version=6.10.0)
```

If I hook the provider in the way recommended by the hardhat docs, then I get the above error when trying to get the signer from hardhat.

```typescript
// See: https://hardhat.org/hardhat-runner/docs/advanced/building-plugins#extending-the-hardhat-provider
extendProvider(async (provider, config, network) => {
  const networkConfig = config.networks[network];
  const { chainId } = networkConfig;
  const rpcUrl = 'url' in networkConfig ? networkConfig.url : '';
  if (chainId) {
    if (!sapphire.NETWORKS[chainId]) return provider;
  } else {
    if (!/sapphire/i.test(rpcUrl)) return provider;

    console.warn(
      'The Hardhat config for the network with `url`',
      rpcUrl,
      'did not specify `chainId`.',
      'The RPC URL looks like it may be Sapphire, so `@oasisprotocol/sapphire-hardhat` has been activated.',
      'You can prevent this from happening by setting a non-Sapphire `chainId`.',
    );
  }
  return sapphire.wrap(provider);
});
```

PR #290, opened 2024-03-19
Remove OpenZeppelin Context

aefhm:
## Description

Our usage of `Context` is minimal.


CedarMist:
Given this was the only thing that required the openzeppelin dependency, surely we can remove that too?

aefhm:
> Given this was the only thing that required the openzeppelin dependency, surely we can remove that too?

We still have `WrappedRose`.
https://github.com/oasisprotocol/sapphire-paratime/blob/36f00be823e9a246f67c99e79e0df721a16faf41/contracts/contracts/WrappedROSE.sol#L4-L5

CedarMist (approved):

PR #268, opened 2024-01-31
integrations/viem: initial package work

CedarMist:
Ideas:

 * Remove ethers dependency from `cipher.ts`
    * Share `cipher.ts` between multiple packages (so, this remains in the core?)
 * Add a `integrations/ethers` package
 * Share `tsconfig` between the packages
 
TODO:

 * [ ] Add `examples/viem`
 * [ ] Add `examples/wagmi`

PR #225, opened 2023-11-17
Add Viem documentation

aefhm:
#### Description

Viem has changed a bit since when we opened #164. This snippet works with Viem [examples](https://github.com/wagmi-dev/viem/tree/main/examples).

#### TODO

- [x] Additional documentation
- [x] File PR for Viem chains
- [x] Simplify import conditional on https://github.com/wevm/viem/pull/1541
- [ ] Add test with `Vigil.sol`

PR #164, opened 2023-08-20
clients/js: add viem compat wrapper

nhynes:

PR #292, opened 2024-03-22
integrations/wagmi: initial re-work of wagmi support

CedarMist:
Closes #268
Closes #164
Closes #225

 * Simplifies `fetchCallDataPublicKey`, Node 18+ support has native `fetch` so no need to import `http` or `https` depending on URL.
 * Bumps `@oasisprotocol/sapphire-paratime` package to 1.3.3
 * Adds 'background mode' for calldata public key fetcher, so it can work with synchronous code
 * Adds `wagmi` example
 * Adds `sapphire-wagmi-v2` integration package
 * Adds `sapphire-viem-v2` integration package
 * Adds `hardhat-viem` example
 * Adds 'ensure everything is encrypted' proxy

Couldn't get [Synpress](https://github.com/Synthetixio/synpress) working with the Wagmi example so have to manually verify.


CedarMist:
So, I'm still running into this problem:

```

PASS test/signed_calls.spec.ts (5.026 s)
PASS test/cipher.spec.ts (5.593 s)
FAIL test/compat.spec.ts (5.674 s)
  ● fetchRuntimePublicKey › non public key provider

    ReferenceError: fetch is not defined

      94 |       `Unable to fetch runtime public key for network with unknown ID: ${chainId}.`,
      95 |     );
    > 96 |   const fetchImpl = (opts?.fetch ?? globalThis?.fetch) ?? fetch;
         |                                                           ^
      97 |   if (!fetchImpl) {
      98 |     throw new Error('No fetch implementation found!');
      99 |   }
```

This works fine with Node  v18.19.0, the CI tests say it's running Node v18.19.1:

```
Run actions/setup-node@v4
  with:
    node-version: 18
    always-auth: false
    check-latest: false
    token: ***
  env:
    OASIS_UNSAFE_SKIP_AVR_VERIFY: 1
    OASIS_UNSAFE_KM_POLICY_KEYS: 1
    OASIS_UNSAFE_ALLOW_DEBUG_ENCLAVES: 1
Found in cache @ /opt/hostedtoolcache/node/18.19.1/x64
Environment details
  node: v18.19.1
  npm: 10.2.4
  yarn: 1.22.22
```

However, this would indicate that the version of Node being used isn't the v18.19.1 that's reported, as the tests pass locally and `fetch` is defined.

CedarMist:
There are some remaining problems with packaging, as seen by the following imports from `examples/wagmi`.

image

Because the `sapphireTransport` function uses `sapphire.wrap(url)` it imports most of Ethers v6.

Secondly, it seems two versions of `viem` and `@wagmi/core` are being imported:

 * @wagmi+core@2.6.9_@types+react@18.2.67_react@18.2.0_typescript@5.4.3_viem@2.8.16/node_modules/@wagmi/core/dist/esm
 * @wagmi+core@2.6.9_react@18.2.0_typescript@5.4.3_viem@2.8.18/node_modules/@wagmi/core/dist/esm
 * viem@2.8.18_typescript@5.4.3/node_modules/viem/_esm
 * viem@2.8.16_typescript@5.4.3/node_modules/viem/_esm

Adding `viem=2.x` as a peer dependency to `integrations/wagmi` hasn't helped

CedarMist:
Solving the `imports all of Ethers v6` problem will reduce the bundle size by up to 170kb.

The reason Ethers is imported is because `sapphire.wrap('https://gateway...')` wraps an Ethers v6 `JsonRpcApiProvider`, which is only needed because the Viem Transport requires an EIP-1193 compatible connector and there's no way to directly hook the `http` transport.

It's possible to intercept responses via `onFetchResponse`: https://github.com/wevm/viem/blob/5a015273a8868efb82640005c0bb18acfa134c1b/src/clients/transports/http.ts#L42

However, there may be another way, by wrapping the `.request` method returned by `createTransport` which invokes `buildRequest`

CedarMist:
I've discovered a problem, which I had noted before in calldatapublickey.ts

    // XXX: if provider switch chain, may return cached key for wrong chain

If I switch chains, with the Wagmi example, the calldata public key returned is for the wrong chain.

docs/browser.md
```diff
@@ -65,11 +71,11 @@ the following changes:
 --- a/hardhat-boilerplate/frontend/src/components/Dapp.js
 +++ b/hardhat-boilerplate/frontend/src/components/Dapp.js
 @@ -2,6 +2,7 @@
- 
+
  // We'll use ethers to interact with the Ethereum network and our contract
  import { ethers } from "ethers";
```
lubej, line 76: To try to minimize the bundle. I would try to avoid this import, which is probably the reason why the whole library is included. I did look at ethers-v6, and it looks like they have ESM build, which should enable tree shaking.

What I suggest, if we are only using only JsonRpcProvider, to import it like this:
```typescript
import { JsonRpcProvider } from 'ethers/providers'
```

And see if it affects the bundle. I am not sure if you can do even more specific imports, so that not all providers all imported.

.github/workflows/ci-test.yaml
```diff
@@ -114,8 +114,6 @@ jobs:
             clients/js/lib
 
       - name: Test JS client
-        env:
-          NODE_OPTIONS: "--no-experimental-fetch" # https://github.com/nock/nock/issues/2397
```
lubej, line 118: Was `nock` and `node-fetch` removed from dependencies, as this was clearly there, to bypass node fetch, in order for `nock` to work?

clients/js/README.md
```diff
@@ -28,6 +28,33 @@ import '@oasisprotocol/sapphire-hardhat';
...
+    [sapphireLocalnet.id]: sapphireTransport(),
+  },
+});
+```
+
+Please note that [EIP-6963] (Multi Injected Provider Discovery) is not currently
+supported with Wagmi.
+
+[EIP-6963]: https://eips.ethereum.org/EIPS/eip-6963
+
```
lubej, line 57: What? https://wagmi.sh/react/api/createConfig#multiinjectedproviderdiscovery

So `@oasisprotocol/sapphire-paratime` does not support `EIP-6963`? This reads like `Wagmi` does not support `EIP-6963`. Can we change the wording here a bit? So it's more clear.

clients/js/src/compat.ts
```diff
@@ -247,6 +252,26 @@ export function wrapEthersProvider<P extends Provider | Ethers5Provider>(
...
+      method: string;
+      params: any[];
+    }) {
+      const cipher = await filled_options.fetcher.cipher(provider);
+      const { method, params } = await prepareRequest(
+        args,
+        cipher,
+        signer as any,
+      );
+      const res = await (provider as any).send(method, params ?? []);
```
lubej, line 267: Please avoid `any`.

docs/browser.md
lubej: Ah, my bad, this is in `.md`. I meant in our code 😅

docs/guide.mdx
```diff
@@ -119,12 +119,38 @@ it'll work just fine.
...
+Sapphire is programmable using any language that targets the EVM, such as Solidity,
+Fe or Vyper. If you prefer to use an Ethereum framework like Hardhat or Foundry,
+you can also use those with Sapphire; all you need to do is set your Web3 gateway URL.
 You can find the details of the Oasis Sapphire Web3 endpoints
 [here](https://github.com/oasisprotocol/docs/blob/main/docs/dapp/sapphire/README.mdx#rpc-endpoints).
 
+
+#### Wagmi
+
+Sapphire will mostly work with Wagmi v2 or Viem v2 out of the box, however in
```
lubej, line 131: Can we avoid "`mostly`" word. And try to explain why?

examples/wagmi/src/App.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const { disconnect } = useDisconnect();
+  const { data: walletClient } =  useWalletClient();
+  const [deployHash, setDeployHash] = useState<undefined | `0x${string}`>();
+  const [contractAddress, setContractAddress] = useState<undefined | `0x${string}`>();
+  const [writeTxHash, setWriteTxHash] = useState<undefined | `0x${string}`>();
+  const [readResult, setReadResult] = useState<bigint|undefined>();
+  const publicClient = usePublicClient();
+  const {
+    data:deployReceipt,
+    error:deployTxError,
```
lubej, line 47: Do we have `prettier` in this repo? This doesn't looks correctly formatted to me. I would expect a space after `:`

examples/wagmi/src/main.tsx
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react'
+import ReactDOM from 'react-dom/client'
+import { WagmiProvider } from 'wagmi'
+
+import App from './App.tsx'
+import { config } from './wagmi.ts'
+
+import './index.css'
+
+//globalThis.Buffer = Buffer
```
lubej, line 12: Probably remove this? And the import above? Since it is not used.

examples/wagmi/tests/e2e/specs/tests.js
```diff
@@ -0,0 +1,36 @@
...
+      cy.get('#connect-injected-sapphire').click();
+      console.log('Accept Metamask');
+      cy.acceptMetamaskAccess();
+      console.log('Checking...');
+      cy.get('#accountAddress').should(
+        'have.text',
+        '0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266',
+      );
+    });
+    */
```
lubej, line 20: Dead code. You can remove it with commit, so it stays in history. But I wouldn't commit this to the repo.

integrations/wagmi/src/index.ts
```diff
@@ -0,0 +1,80 @@
+import { wrapEIP1193Provider, wrap } from '@oasisprotocol/sapphire-paratime';
+import type { Transport } from '@wagmi/core';
+import { injected, custom } from '@wagmi/core';
+
+type Window = {
+  ethereum?: any;
```
lubej, line 6: Avoid `any`

lubej: For that matter, avoid any everywhere in this file.

docs/browser.md
```diff
@@ -78,16 +84,16 @@ the following changes:
  // to use when deploying to other networks.
 -const HARDHAT_NETWORK_ID = '1337';
 +const HARDHAT_NETWORK_ID = '23295'; // Sapphire Testnet
```
lubej, line 86: `SAPPHIRE_TESTNET_NETWORK_ID`?

docs/guide.mdx
```diff
@@ -119,12 +119,38 @@ it'll work just fine.
 
 ### Languages & Frameworks
 
-Sapphire is programmable using any language that targets the EVM, such as Solidity
-or Vyper. If you prefer to use an Ethereum framework like Hardhat or Foundry, you
-can also use those with Sapphire; all you need to do is set your Web3 gateway URL.
+Sapphire is programmable using any language that targets the EVM, such as Solidity,
```
matevz, line 122: Let's move the *Wallets* and *Languages & Frameworks* under the *Integrating Sapphire* section.

docs/guide.mdx
```diff
@@ -119,12 +119,38 @@ it'll work just fine.
...
+Fe or Vyper. If you prefer to use an Ethereum framework like Hardhat or Foundry,
+you can also use those with Sapphire; all you need to do is set your Web3 gateway URL.
 You can find the details of the Oasis Sapphire Web3 endpoints
 [here](https://github.com/oasisprotocol/docs/blob/main/docs/dapp/sapphire/README.mdx#rpc-endpoints).
 
+
+#### Wagmi
```
matevz, line 129: This chapter is more about writing contracts, deploying and managing them. And it is already getting pretty long. Wagmi is a react-based framework right? I would move it to the [Browser](https://deploy-preview-292--oasisprotocol-sapphire-paratime.netlify.app/dapp/sapphire/browser) chapter instead and split the chapter into *Hardhat* (most of the existing content) and *Wagmi* (this section).

examples/wagmi/src/wagmi.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  testnet: true,
+});
+
+export const config = createConfig({
+  multiInjectedProviderDiscovery: false,
+  chains: [
+    sapphireTestnet,
+    sapphireLocalnet
+  ],
```
matevz, line 25: Mainnet?

```suggestion
    sapphire
  ],
```

examples/wagmi/src/App.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const {
+    data:writeReceipt,
+    error:writeTxError,
+  } = useWaitForTransactionReceipt({hash:writeTxHash, confirmations: 1});
+
+  async function doDeploy () {
+    const hash = await walletClient?.deployContract({
+      abi: StorageABI,
+      bytecode: '0x608060405234801561000f575f80fd5b506101438061001d5f395ff3fe608060405234801561000f575f80fd5b5060043610610034575f3560e01c80632e64cec1146100385780636057361d14610056575b5f80fd5b610040610072565b60405161004d919061009b565b60405180910390f35b610070600480360381019061006b91906100e2565b61007a565b005b5f8054905090565b805f8190555050565b5f819050919050565b61009581610083565b82525050565b5f6020820190506100ae5f83018461008c565b92915050565b5f80fd5b6100c181610083565b81146100cb575f80fd5b50565b5f813590506100dc816100b8565b92915050565b5f602082840312156100f7576100f66100b4565b5b5f610104848285016100ce565b9150509291505056fea26469706673582212201bc715d5ea5b4244a667a55f9fd36929a52a02208d9b458fdf543f5495011b2164736f6c63430008180033',
```
matevz, line 58: Which contract is this?

matevz (changes requested):

clients/js/README.md
CedarMist: EIP-6963 is supported by Wagmi, however due to the way that MIPD is implemented in Wagmi I'm unable to find a way of wrapping the detected providers with Sapphire.

docs/browser.md
CedarMist: Ethers does support tree shaking, and in the sapphire-paratime package we're importing only specific functions/classes.

The only improvement that can be made is removing the `wrap` function in favor of `wrapEthersSigner` and `wrapEthersProvider` etc.

There are a few places where Ethers is being used lazily where an EIP-1193 compatible provider could be used directly. Combine that with removing `wrap` and tree-shaking will be better able to exclude unused code.

.github/workflows/ci-test.yaml
CedarMist: After removal of the code which conditionally imports `http` or `https` modules Nock seems to be working - with node v18 which has built-in `fetch` support.

We only support node v18+

docs/browser.md
CedarMist: This relates to a specific tutorial which modifies the hardhat boilerplate project to use Sapphire. Modifying that is out of the scope of this PR.

.github/workflows/ci-test.yaml
lubej: I am referring to why are `nock` and `nock-fetch` still listed in dependencies, if we are now using only node:

https://github.com/oasisprotocol/sapphire-paratime/blob/2155379b2207aefb9bbf9d2f6850be4601e7bb64/clients/js/package.json#L62-L63

docs/browser.md
lubej: Not sure, why https://github.com/oasisprotocol/sapphire-paratime/pull/292#issuecomment-2016555236 the ethers chunk is like 1/3th of the bundle. But ok I guess. I just wanted to point out, we could try with more specific imports, to see if it optimizes the bundle.

clients/js/README.md
lubej: I would just make it more clear that `@oasisprotocol/sapphire-paratime` does not support Wagmi's EIP-6963 implementation and you need to make a config with `multiInjectedProviderDiscovery` set to `false` in the dApp.

integrations/wagmi/src/index.ts
CedarMist: This is unavoidable in only one instance, which remains.

Because the `viem` EIP1193Provider type is so specific it's incompatible with any other type definition of an EIP1193 provider.

examples/wagmi/src/App.tsx
CedarMist: I'm using `biome`, have re-formatted using it.

docs/browser.md
CedarMist: Yup I'm very aware that it's a large part of the bundle, but I won't be able to address that in this PR without making some major changes to sapphire-paratime package which requires introducing a sapphire-ethers-v6 package.

integrations/wagmi/src/index.ts
```diff
@@ -39,16 +40,16 @@ export function injectedWithSapphire(): ReturnType<typeof injected> {
             // Note: providers are cached as connectors are frequently retrieved
             //       it prevents sapphire wrapper being called multiple times
             //       which spams the RPC with oasis_callDataPublicKey requests
-            if (!(window.ethereum in cachedProviders)) {
-              cachedProviders[window.ethereum] = wrapEIP1193Provider(
-                window.ethereum,
-              );
+            if( ! (cachedProviders.has(window.ethereum)) ) {
+              const wp = wrapEIP1193Provider(window.ethereum as any) as EIP1193Provider & SapphireAnnex;
```
lubej, line 44: If you want to cast this to `EIP1193Provider & SapphireAnnex`, please do it the following way:
```typescript
const wp = wrapEIP1193Provider(window.ethereum as unknown as EIP1193Provider);
```

Please cast to `unknown` first and then to desired type. `any` messes up the type completely.

integrations/wagmi/src/index.ts
lubej: Looks good now 👍

examples/wagmi/src/App.tsx
CedarMist: I've added a comment containing the Solidity source above the bytecode

clients/js/README.md
CedarMist: This has been addressed in the README.md file from `integrations/wagmi-v2`

docs/guide.mdx
CedarMist: This has been removed from the guide, it's addressed in the wagmi README

integrations/wagmi-v2/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @returns Same as injected()
+ */
+export function injectedWithSapphire(): ReturnType<typeof injected> {
+	return injected({
+		target: () => {
+			return {
+				id: "injected-sapphire",
+				name: "Injected (Sapphire)",
+				provider(window?: Window) {
```
lubej, line 42: Is this window parameter necessary? Is it the expected function signature?

lubej (dismissed):
LGTM overall

clients/js/scripts/proxy.ts
```diff
@@ -43,7 +43,11 @@ interface JSONRPCRequest {
 
 async function onRequest(req: IncomingMessage, response: ServerResponse) {
   if (req.method !== 'POST') {
-    response.writeHead(500, 'Not POST!');
+    // An initial prefetch request will be made to determine if CORS is allowed.
```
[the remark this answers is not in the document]
lubej: I guess it should be, initial prefetch, to see if POST is allowed, or something? 🤷

integrations/wagmi-v2/src/index.ts
CedarMist: Yes, it is expected:

https://github.com/wevm/wagmi/blob/8e9ee793c4f378de523239bb8db63cbf09cc980e/packages/core/src/connectors/injected.ts#L472

clients/js/scripts/proxy.ts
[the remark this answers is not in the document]
CedarMist: Not sure, I had to fudge this to get the proxy working as an RPC provider with `localhost`

CedarMist:
One last thing I really should check is hardhat support with wagmi/viem:

Wagmi supports hardhat like this: https://wagmi.sh/cli/api/plugins/hardhat
Hardhat supports Wagmi like this: https://www.npmjs.com/package/@nomicfoundation/hardhat-viem

clients/js/scripts/proxy.ts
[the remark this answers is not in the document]
lubej: Ok I guess, comment wasn't clear to me

CedarMist:
Currently it doesn't seem possible to wrap an existing client and override its `transport` or the `transport.request` property, as that's getting bound somewhere in a closure.

So instead I considered wrapping an arbitrary transport, as such:

```typescript
export function wrapTransport<T extends Transport>(t:T) {
	return (params:any) => {
		const u = t(params);
		if( Reflect.get(u, SAPPHIRE_WRAPPED_VIEM_TRANSPORT) ) {
			return u;
		}
		const p = wrapEIP1193Provider(u);
		Reflect.set(p, SAPPHIRE_WRAPPED_VIEM_TRANSPORT, true);
		return p;
	};
}
```

But that errors with the following due to the user of Ethers BrowserProvider in `compat.ts` to detect whether or not it has a signer attached. However using the `sapphireTransport` works as expected.

```
    Error: could not coalesce error (error={ "code": -32601, "payload": { "method": "eth_requestAccounts", "params": [  ] } }, payload={ "method": "eth_requestAccounts", "params": [  ] }, code=UNKNOWN_ERROR, version=6.10.0)
      at makeError (/src/sapphire-paratime/node_modules/.pnpm/ethers@6.10.0/node_modules/ethers/src.ts/utils/errors.ts:694:21)
      at BrowserProvider.getRpcError (/src/sapphire-paratime/node_modules/.pnpm/ethers@6.10.0/node_modules/ethers/src.ts/providers/provider-jsonrpc.ts:1049:25)
      at BrowserProvider.getRpcError (/src/sapphire-paratime/node_modules/.pnpm/ethers@6.10.0/node_modules/ethers/src.ts/providers/provider-browser.ts:107:22)
      at BrowserProvider.getSigner (/src/sapphire-paratime/node_modules/.pnpm/ethers@6.10.0/node_modules/ethers/src.ts/providers/provider-browser.ts:136:28)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

integrations/viem-v2/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+	env: {
+		SAPPHIRE_LOCALNET_HTTP_PROXY_PORT?: string;
+	};
+};
+
+/**
+ * This environment variable allows for the sapphire-localnet port to be
+ * overridden via the command-line. This is useful for debugging with a proxy.
+ *
+ * Note: this will fail gracefully in-browser
```
lubej, line 22: How about checking if `window` exists, and not even execute the code, that depends on Node.js runtime?

lubej (commented):
One note, please next time don't create huge PRs, it's a pain to review. I know it's hard, when adding new features, but you could have stacked the PRs on top of each other. Like `wagmi` & `viem` part, could have been easily 2 different PRs. Also an example projects, could have been added separately.

integrations/viem-v2/src/index.ts
CedarMist: I assumed some people may use a `node-process` polyfill

clients/js/src/compat.ts
```diff
@@ -247,6 +252,26 @@ export function wrapEthersProvider<P extends Provider | Ethers5Provider>(
     estimateGas: hookEthersCall(provider, 'estimateGas', filled_options),
   };
 
+  // Provide an EIP-1193 compatible endpoint for Ethers providers
+  if (!('request' in provider) && 'send' in provider) {
```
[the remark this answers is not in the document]
CedarMist: `'request' in provider` is shorter and more explicit than `isEip1193Provider(provider)`

I have a refactor waiting in my head that removes this messiness

clients/js/src/compat.ts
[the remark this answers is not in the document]
CedarMist: It doesn't solve this `any` instance, the `send` interface is exposed by `JsonRpcProvider` and `BrowserProvider`, but this could be from ether the Ethers 5 or Ethers 6 packages.

clients/js/src/calldatapublickey.ts
```diff
@@ -239,6 +222,33 @@ export class KeyFetcher extends AbstractKeyFetcher {
...
+  public async runInBackground(upstream: UpstreamProvider) {
+    if (this.#isBackgroundRunning) return;
+
+    // Reduce wait time by 10%, so it eagerly fetches
+    const waitMilliseconds =
+      this.timeoutMilliseconds - this.timeoutMilliseconds / 10;
+
+    while (this.#isBackgroundRunning) {
+      await new Promise((resolve) =>
+        setTimeout(resolve, this.timeoutMilliseconds),
```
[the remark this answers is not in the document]
CedarMist: Good point, I have refactored the timeout logic

examples/wagmi-v2/src/App.tsx
```diff
@@ -0,0 +1,224 @@
...
+	useConnect,
+	useDisconnect,
+	usePublicClient,
+	useWaitForTransactionReceipt,
+	useWalletClient,
+} from "wagmi";
+import type { Abi } from "abitype";
+
+/*
+// SPDX-License-Identifier: UNLICENSED
```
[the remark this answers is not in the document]
CedarMist: Have changed everything to Apache-2.0 for consistency

lubej:
> > One note, please next time don't create huge PRs, it's a pain to review. I know it's hard, when adding new features, but you could have stacked the PRs on top of each other. Like `wagmi` & `viem` part, could have been easily 2 different PRs. Also the example projects, could have been added separately.
> 
> Have you tested any of the code while reviewing, or just looking at it?

Just looking at it. What is your point? The examples could have been branched upon `sapphire-wagmi-v2` and `sapphire-viem-v2` integrations.

CedarMist:
> Just looking at it. What is your point? The examples could have been branched upon `sapphire-wagmi-v2` and `sapphire-viem-v2` integrations.

The examples are the unit tests

CedarMist:
Fuck it

PR #299, opened 2024-04-02
Update documented addresses

aefhm:
## Description

Add Celer IM message executor.
Add Safe Singleton Factory.

## TODO

- [x] Find out mainnet info
- [x] Update tokens
- [x] Suggest alternative section

### DIFF
--- a/docs/addresses.md
+++ b/docs/addresses.md
@@ -45,10 +45,15 @@ description: List of Standard Contract Addresses
 
 ## Deployments
 
-| Name | Mainnet Address | Testnet Address | Documentation | Source |
-| ---- | --------------- | --------------- | ------------- | ------ |
-| [Celer IM Executor][message-executor] | - | `0x9C850D230FFFaCEf1E2D1741a00080856630e455` | [Testnet][message-executor-testnet] | [Message Executor][message-executor-source] |
+| Name | Mainnet Address | Testnet Address | Source |
+| ---- | --------------- | --------------- | ------ |
+| [Celer IM Executor][message-executor] | Multiple executors available | [`0x9C850D230FFFaCEf1E2D1741a00080856630e455`][message-executor-testnet] | [Message Executor][message-executor-source] |
+| [Celer MessageBus][message-bus] | [`0x9Bb46D5100d2Db4608112026951c9C965b233f4D`][message-bus-mainnet] | [`0x9Bb46D5100d2Db4608112026951c9C965b233f4D`][message-bus-testnet] | [Message bus][message-bus-source] |
 
 [message-executor]: https://im-docs.celer.network/developer/development-guide/message-executor
 [message-executor-source]: https://github.com/celer-network/im-executor
 [message-executor-testnet]: https://explorer.oasis.io/testnet/sapphire/address/0x9C850D230FFFaCEf1E2D1741a00080856630e455
+[message-bus]: https://im-docs.celer.network/developer/development-guide/message-executor
+[message-bus-source]: https://github.com/celer-network/sgn-v2-contracts/blob/6af81b55a13a7aacab9a4d92a38d374d46c0fdbf/contracts/message/messagebus/MessageBus.sol
+[message-bus-mainnet]: https://explorer.oasis.io/mainnet/sapphire/address/0x9Bb46D5100d2Db4608112026951c9C965b233f4D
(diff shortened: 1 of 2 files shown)


docs/addresses.md
```diff
@@ -8,6 +8,7 @@ description: List of Standard Contract Addresses
 |--------------|--------------------------------------------|--------------------------------------------|------------------------------------------------------------------|---------------------------------|
 | [Multicall V3][multicall] | `0xcA11bde05977b3631167028862bE2a173976CA11` | - | [Mainnet][multicall-verify-mainnet] | [Multicall3.sol][multicall-source] |
 | [Wrapped ROSE][wrose-dapp] | `0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3` | `0xB759a0fbc1dA517aF257D5Cf039aB4D86dFB3b94` | [Mainnet][wrose-verify-mainnet], [Testnet][wrose-verify-testnet] | [WrappedROSE.sol][wrose-source] |
+| [Celer IM Executor][message-executor] | - | `0x9C850D230FFFaCEf1E2D1741a00080856630e455` | [Testnet][message-executor-testnet] | [Message Executor][message-executor-source] |
```
matevz, line 11: Where's the Mainnet address?

Also, since this is not a contract address, it doesn't really make sense to put anything into the `Verify` column IMO.

docs/addresses.md
```diff
@@ -18,6 +19,10 @@ description: List of Standard Contract Addresses
 [wrose-verify-mainnet]: https://sourcify.dev/#/lookup/0x8Bc2B030b299964eEfb5e1e0b36991352E56D2D3
 [wrose-verify-testnet]: https://sourcify.dev/#/lookup/0xB759a0fbc1dA517aF257D5Cf039aB4D86dFB3b94
 
+[message-executor]: https://im-docs.celer.network/developer/development-guide/message-executor
+[message-executor-source]: https://github.com/celer-network/im-executor
+[message-executor-testnet]: https://explorer.oasis.io/testnet/sapphire/address/0x9C850D230FFFaCEf1E2D1741a00080856630e455
+
 # Celer cBridge Tokens (Mainnet)
```
matevz, line 26: Please use consistent chapter names, e.g.
- Celer cBridge Tokens (Mainnet)
- Celer cBridge Tokens (Testnet)

This fix should probably go here https://github.com/oasisprotocol/sapphire-paratime/blob/main/docs/_fetch-cbridge-tokens.py#L30?

matevz (changes requested):

docs/addresses.md
aefhm: > Where's the Mainnet address?

Fair. Looking.

> Also, since this is not a contract address, it doesn't really make sense to put anything into the Verify column IMO.

Makes sense. I like having a unified page for configurations. I'll suggest a renaming of this page and then add another appropriate section.

docs/addresses.md
aefhm: > Please use consistent chapter names, e.g.

👍 

> This fix should probably go here https://github.com/oasisprotocol/sapphire-paratime/blob/main/docs/_fetch-cbridge-tokens.py#L30?

I agree with your previous [comment](https://github.com/oasisprotocol/sapphire-paratime/pull/299#discussion_r1549277903) that these aren't tokens, so will suggest another section for them.

Separately, the list of tokens are out of date, and I'll patch that.

aefhm: > There are multiple mainnet executors, one for each project, but we could monitor the messagebus contract address
For testnet it's wise to monitor both executor and the messagebus contract address - as people may run their own executor

from @CedarMist

aefhm:
Pulling back from history https://github.com/safe-global/safe-singleton-factory/pull/78/files

aefhm:
@matevz one more look? I think changes requested are there now.

CedarMist (approved):
Everything seems correct

The 'Contract Addresses and Deployments' header does seem a bit long, but that's not a blocker.

aefhm:
> Everything seems correct
> 
> The 'Contract Addresses and Deployments' header does seem a bit long, but that's not a blocker.

Oh yeah. Hmm, maybe we should just have 2 total sections on this page "Addresses" with two sections 1 for "Contracts" and another for the bridged tokens instead.

CedarMist:
Am happy to merge and then look at the docs from a higher-level perspective afterwards, having all the standard addresses on one page is good and easy to read.

PR #303, opened 2024-04-24
Add Viem v2 & Wagmi v2 packages, and remove Ethers dependency

CedarMist:
## Description

Add support for Viem v2 and Wagmi v2. Remove Ethers dependency.

Replaces https://github.com/oasisprotocol/sapphire-paratime/pull/164

### Details

* Simplified `sapphire-paratime` dependencies
   * Removed `ethers`, this reduces bundled size significantly when used standalone or with Viem
   * Removed `tweetnacl.js`, extracted necessary functions into `munacl.ts` 
   * Removed generic `sapphire.wrap` function, it now exposes `wrapEthereumProvider` which only wraps EIP-1193 compatible providers

* Tests:
  * Removed many mock based tests, instead relying more heavily on integration tests to verify actual functionality
  * All CI tests run through a local RPC proxy (`clients/js/scripts/proxy.ts`) which will die if any unencrypted transactions, calls or gas estimates are made.
    * This ensures that the clients & integrations apply encryption consistently, and will prevent any regressions where encryption is accidentally broken.

* Integrations:
  * Each SDK has its own integration package + tests
    * `@oasisprotocol/sapphire-ethers-v6`
    * `@oasisprotocol/wagmi-v2`
    * `@oasisprotocol/viem-v2`

* Removed  duplicate example projects (e.g. where two projects use the exact same underlying libraries)
  * Each example project also serves as an integration test
    * `hardhat-viem` (Hardhat + Viem)
    * `onchain-signer` (Hardhat + Ignition + Ethers v6)
    * `hardhat-boilerplate` (Hardhat + Ethers v5 + frontend & bundling)
    * `wagmi-v2` (Wagmi + frontend & bundling)
    * `wagmi-v1`


CedarMist:
Feedback:
 - use @noble/curves rather than munacl.ts

clients/js/src/cipher.ts
```diff
@@ -1,168 +1,200 @@
...
-    return cbor.encode({ [prop]: { nonce, data: ciphertext } });
+    return asBytesLike(calldata, cborEncode(envelope));
   }
 
-  /** Decrypts the data contained within a hex-encoded serialized envelope. */
-  public async decryptEncoded(callResult: BytesLike): Promise<string> {
-    return hexlify(
-      await this.decryptCallResult(cbor.decode(getBytes(callResult))),
-    );
+  public decryptCall(envelopeBytes: BytesLike): BytesLike {
```
aefhm, line 122: Test only function or do we expect developers to use this while debugging etc?

clients/js/test/cipher.spec.ts
```diff
@@ -61,54 +26,22 @@ describe('X25519DeoxysII', () => {
...
-      await cipher.decryptCallResult({
-        unknown: {
-          nonce: resData.nonce,
-          data: resData.ciphertext,
-        },
-      }),
-    ).toEqual(DATA);
+    for (let i = 1; i < 512; i += 30) {
+      const expected = nacl.randomBytes(i);
+      const decoded = cipher.decryptCall(cipher.encryptCall(expected));
```
aefhm, line 33: (nit) potentially test this separately?

aefhm (commented):
Comments cover client code. Will send notes on integrations next. Still need to QA as well.

.github/workflows/contracts-test.yaml
```diff
@@ -21,6 +21,8 @@ jobs:
           --rm
           --health-cmd="/oasis-node debug control wait-ready -a unix:/serverdir/node/net-runner/network/client-0/internal.sock"
           --health-start-period=90s
+    env:
+      SAPPHIRE_LOCALNET_HTTP_PROXY_PORT: 3001
```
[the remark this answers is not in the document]
CedarMist: Matevz suggested setting the localnet container to be 'Encrypted TXs Only' via [another PR](https://github.com/oasisprotocol/oasis-web3-gateway/pull/340)

However, whether or not we want the TX to be encrypted is more nuanced, I only want encrypted transactions enforced for a couple of accounts, and I want the ability to selectively turn on/off the enforcement on a per-test level (meaning, we can unset `SAPPHIRE_LOCALNET_HTTP_PROXY_PORT` for a test if needed).

Using the proxy approach also means there's easier JSON-RPC request logging, so if there's a failure we can see the last handful of requests that lead upto it.

clients/js/package.json
```diff
@@ -35,35 +35,30 @@
...
     "eslint": "^8.23.1",
     "eslint-config-prettier": "^8.5.0",
+    "ethers": "6.x",
     "jest": "^29.7.0",
-    "nock": "^13.4.0",
-    "node-fetch": "^2.6.7",
     "prettier": "^2.7.1",
     "ts-jest": "^29.1.1",
     "ts-node": "^10.9.2",
+    "tweetnacl": "^1.0.3",
```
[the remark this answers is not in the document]
CedarMist: This is only used in `cipher.spec.ts` and `utils.ts`, this could be switched to `@noble/curves`.

The bug in question:
> IMPORTANT BUG FIX. Due to a bug in calculating carry in
modulo reduction that used bit operations on integers larger than
32 bits, nacl.sign or nacl.sign.detached could have created
incorrect signatures.
> 
> This only affects signing, not verification.

matevz:
>     * tag current branch as 1.x / stable
>     * new stuff goes as 2.x / beta?

Sleeping over this, I would name the stable branch as `clients/js/1.3.x` and keep the new refactored codebase inside the `main` branch. If we find anything worth fixing in the stable branch as well, then we backport the fix from `main` to `clients/js/1.3.x`. If the fix is solely for 1.3.x, then we just fix the issue on the branch directly.

clients/js/src/munacl.ts
```diff
@@ -0,0 +1,654 @@
+// SPDX-License-Identifier: Unlicense
+
+// Minimum necessary functions extracted and made TypeScript compatible from:
+// https://github.com/dchest/tweetnacl-js/blob/fecde6ecf0eb81e31d54ca0509531ab1b825f490/nacl-fast.js
```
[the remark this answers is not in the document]
CedarMist: Kinda, tweetnacl is still in the development dependencies, and we're only using a very small and well defined subset of the library.

Alternative is move to `@noble/curves` package.

clients/js/scripts/proxy.ts
```diff
@@ -1,36 +1,37 @@
 import { createServer, IncomingMessage, ServerResponse } from 'node:http';
 
-import { decodeRlp, getBytes } from 'ethers';
+import { formatEther, getBytes, toBigInt, Transaction } from 'ethers';
```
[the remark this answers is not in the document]
CedarMist: I'm happy keeping `ethers` as a dev dependency.

clients/js/src/cipher.ts
CedarMist: At the moment it's only used for tests, to ensure roundtrip of encryption/decryption, we also don't use the `encryptResult` outside of tests.

It's not very useful for end-user devs while debugging, as the ephemeral keypair used for encrypting the transactions gets discarded or is otherwise inaccessible.

I guess we could move this into a test utils file?

clients/js/src/provider.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    return res;
+  };
+
+  return makeTaggedProxyObject(f, SAPPHIRE_EIP1193_REQUESTFN, filled_options);
+}
+
+// -----------------------------------------------------------------------------
+
+export function makeTaggedProxyObject<T extends object>(
```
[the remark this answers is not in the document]
CedarMist: I'm going to leave it exported for now until I find a cleaner way.

clients/js/package.json
```diff
@@ -35,35 +35,30 @@
...
   },
   "devDependencies": {
     "@types/jest": "^29.5.11",
     "@types/node": "^18.7.18",
-    "@types/node-fetch": "^2.6.2",
     "@typescript-eslint/eslint-plugin": "^5.38.0",
     "@typescript-eslint/parser": "^5.38.0",
     "eslint": "^8.23.1",
     "eslint-config-prettier": "^8.5.0",
+    "ethers": "6.x",
```
matevz, line 56: Leftover?

aefhm (commented):
I took a QA pass around examples and the Hardhat fork which worked great :)

integrations/wagmi-v2/package.json
```diff
@@ -0,0 +1,70 @@
...
+	"typings": "./dist/_types/index.d.ts",
+	"exports": {
+		"node": {
+			"import": "./dist/_esm/index.js",
+			"types": "./dist/_types/index.d.ts"
+		},
+		"types": "./dist/_types/index.d.ts",
+		"default": "./dist/_esm/index.js"
+	},
+	"scripts": {
```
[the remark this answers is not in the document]
CedarMist: We don't have automated tests for this.

Will need Synpress / DappWrite tests for the frontend related stuff as can't test Wagmi + MetaMask without that.

clients/js/package.json
CedarMist: Yup it's in dev dependencies, it's used for various tests & proxy server stuff. It doesn't affect the released package.

integrations/ethers-v6/src/index.ts
```diff
@@ -0,0 +1,261 @@
...
+			// Calls can be unsigned, but must be enveloped.
+			call: hookEthersCall(provider, "call", filled_options, request),
+			estimateGas: hookEthersCall(
+				provider,
+				"estimateGas",
+				filled_options,
+				request,
+			),
+			request,
+			// TODO: wrap `getSigner()`
```
aefhm, line 206: Or I guess devs could do this themselves.

aefhm (approved):
I vote to merge, and we can iterate from there.

integrations/ethers-v6/src/index.ts
CedarMist: So, I can't easily test this without a provider like MetaMask, as `getSigner` will try to enumerate accounts, e.g. in geth.

matevz (approved):
I tried the SIWE mechanism as an alternative to the signed queries that were removed in this PR. I can confirm the alternative is viable, so feel free to merge this.

PR #304, opened 2024-04-25
Hardhat example comment issue

Mylanos:
The comment with a command that hints on the way how the script file should be used references some other file instead of the script file the comment is contained in.


aefhm (approved):
Thanks!

PR #315, opened 2024-05-15
Fix Go client contract deployment

aefhm:
## Description

`PackSignedCall` is part of the deployment flow and needs to support the case where `msg.To` is `nil`.

https://github.com/oasisprotocol/sapphire-paratime/blob/5d6b79e859283f4d707fa41015f33b94dffb78dd/clients/go/compat.go#L90-L97

## TODO

- [ ] Publish new Go client.


kostko (approved):

clients/go/compat.go
```diff
@@ -94,7 +94,12 @@ func PackSignedCall(msg ethereum.CallMsg, cipher Cipher, sign SignerFn, chainID
 	if msg.GasPrice == nil {
 		msg.GasPrice = big.NewInt(DefaultGasPrice) // Must be non-zero for signed calls.
 	}
-	dataPack, err := evm.NewSignedCallDataPack(rsvSigner{sign}, chainID.Uint64(), msg.From[:], msg.To[:], msg.Gas, msg.GasPrice, msg.Value, msg.Data, *leash)
+	// msg.To is nil when deploying
```
matevz, line 97:
```suggestion
	// msg.To is nil when deploying.
```

matevz (approved):
The app crashes when you want to deploy a contract with an encrypted transaction right? Because we're using this go-client library in network-monitoring and we didn't encounter any crashes, but that's probably because we use the unwrapped version of the eth client for contract deployment.

Can you also please fix the linting issues while you're at it.

aefhm:
> The app crashes when you want to deploy a contract with an encrypted transaction right?

I'm using [plain](https://explorer.oasis.io/testnet/sapphire/tx/0x953f2ccb58e07bce8579f6d529c57486128d8575183f2b9a4b9f91f2a12ca860). 🤔 

> probably because we use the unwrapped version of the eth client for the contract deployment.

Which version of the client?

aefhm:
I think my QA issues are independent of this client patch.

aefhm:
@kostko new release to `clients/go/v0.11.2` please?

kostko:
Done.

PR #325, opened 2024-07-01
contracts: fix CBOR unsigned int decoding in Subcall.sol

### LINKED ISSUE #323: `_parseCBORUint` failing in `_decodeReceiptUndelegateStart`
The `consensusTakeReceiptUndelegateStart` function in `Subcall.sol` is failing on testnet with the `InvalidLength` error.

CedarMist:
fixes #323

### DIFF
--- a/contracts/contracts/Subcall.sol
+++ b/contracts/contracts/Subcall.sol
@@ -194,7 +194,7 @@ library Subcall {
         if (prefix <= 0x17) {
             return (offset + 1, prefix);
         }
-        // byte array, parsed as a big-endian integer
+        // Byte array, parsed as a big-endian integer.
         else if (prefix & 0x40 == 0x40) {
             len = uint8(result[offset++]) ^ 0x40;
         }
--- a/contracts/contracts/Subcall.sol
+++ b/contracts/contracts/Subcall.sol
@@ -299,9 +299,6 @@ library Subcall {
                 (offset, endReceipt) = _parseCBORUint64(result, offset);
 
                 hasReceipt = true;
-            } else {
-                // TODO: skip unknown keys & values? For forward compatibility
-                revert InvalidKey();
             }
         }
 
@@ -328,9 +325,6 @@ library Subcall {
                 (offset, amount) = _parseCBORUint128(result, offset);
 
                 hasAmount = true;
-            } else {
-                // TODO: skip unknown keys & values? For forward compatibility
-                revert InvalidKey();
             }
         }
 
--- a/contracts/contracts/Subcall.sol
+++ b/contracts/contracts/Subcall.sol
@@ -199,6 +199,11 @@ library Subcall {
         if (prefix <= 0x17) {
             return (offset + 1, prefix);
         }
+        // Byte array(uint256), parsed as a big-endian integer.
+        else if (prefix == 0x58) {
+            len = uint8(result[++offset]);
+            offset++;
+        }
         // Byte array, parsed as a big-endian integer.
         else if (prefix & 0x40 == 0x40) {
             len = uint8(result[offset++]) ^ 0x40;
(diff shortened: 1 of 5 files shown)


CedarMist:
Ok, minor last thing.

I need to make the subcall stuff not error out if it encounters and unknown key.

contracts/contracts/Subcall.sol
```diff
@@ -182,11 +188,37 @@ library Subcall {
...
     {
-        if (result[offset] & 0x40 != 0x40) revert InvalidLength();
+        uint8 prefix = uint8(result[offset]);
+        uint256 len;
 
-        uint256 len = uint8(result[offset++]) ^ 0x40;
+        if (prefix <= 0x17) {
+            return (offset + 1, prefix);
+        }
+        // byte array, parsed as a big-endian integer
```
matevz, line 197:
```suggestion
        // Byte array, parsed as a big-endian integer.
```

contracts/contracts/Subcall.sol
```diff
@@ -182,11 +188,37 @@ library Subcall {
...
+        else if (prefix & 0x40 == 0x40) {
+            len = uint8(result[offset++]) ^ 0x40;
+        }
+        // unsigned integer, CBOR encoded
```
matevz, line 201:
```suggestion
        // Unsigned integer, CBOR encoded.
```

contracts/contracts/Subcall.sol
```diff
@@ -63,6 +63,12 @@ library Subcall {
     /// CBOR parser expected a key, but it was not found in the map!
     error MissingKey();
 
+    /// Valid cannot be parsed as a uint
```
matevz, line 66: ?

matevz (approved):

matevz (commented):
FYI this section https://www.rfc-editor.org/rfc/rfc8949.html#name-tag-validity mentions that the 128- and 256-bit integers are encoded as 0x40 (byte arrays).

matevz:
> FYI this section https://www.rfc-editor.org/rfc/rfc8949.html#name-tag-validity mentions that the 128- and 256-bit integers are encoded as 0x40 (byte arrays).

Having the _cborParseUInt() split into 64-bit and smaller uints and 128-bit and greater uints codepaths opens potential gas attacks? Should we use gas padding to mitigate this?

CedarMist:
> > FYI this section https://www.rfc-editor.org/rfc/rfc8949.html#name-tag-validity mentions that the 128- and 256-bit integers are encoded as 0x40 (byte arrays).
> 
> Having the _cborParseUInt() split into 64-bit and smaller uints and 128-bit and greater uints codepaths opens potential gas attacks? Should we use gas padding to mitigate this?

I'd prefer if the caller of the receipt parser does the gas padding if it's deemed necessary rather than building it into the parser itself. But yes, I'll add a note in the code about it.

contracts/contracts/Subcall.sol
```diff
@@ -182,11 +188,39 @@ library Subcall {
...
+                len = 1;
+            } else if (prefix == 0x19) {
+                len = 2;
+            } else if (prefix == 0x1a) {
+                len = 4;
+            } else if (prefix == 0x1b) {
+                len = 8;
+            } else if (prefix == 0x1d) {
+                len = 16;
+            } else {
```
lubej, line 213: I see now, that this is actually wrong https://www.rfc-editor.org/rfc/rfc8949.html#name-jump-table-for-initial-byte

contracts/contracts/Subcall.sol
```diff
@@ -170,6 +176,11 @@ library Subcall {
             (uint64, bytes)
         );
 
+        // 0xf6 = null, returns null in case receiptId not found
```
lubej, line 179: Also returns null(`0xf6`) in case epoch in not reached yet.

PR #332, opened 2024-07-12
sapphire-contracts: Add EIP-4361 SIWE parser and base contract

### LINKED ISSUE #327: sapphire-contracts: Add SIWE parser
In light of #303 and removal of the signed queries, add support for on-chain SIWE parsing.

Related oasisprotocol/demo-starter#16

matevz:
Fixes #327 

This PR:
- adds `contracts/SiweParser.sol`, an on-chain parser for SIWE messages
- adds `contracts/auth/A13e.sol`, an abstract interface for authenticated queries (an alternative to signed queries which will get deprecated in https://github.com/oasisprotocol/sapphire-paratime/pull/303)
- adds `contracts/auth/SiweAuth.sol`, an on-chain authenticator for SIWE.
- tests, natspec docs + example

TODO:
- [ ] ~make SiweParser self-contained (merge DateTime into SiweParser?)~ (this doesn't actually improve anything, just adds clutter)
- [x] use error constants in SiweParser and SiweAuth

### Before (signed queries)

```solidity
contract MyContract {
  address private _owner;

  modifier onlyOwner() {
    if (msg.sender != _owner) {
      revert("not allowed");
    }
    _;
  }

  constructor() {
    _owner = msg.sender;
  }

  function getSecretMessage() external view onlyOwner returns (string memory) {
    return "Very secret message";
  }
}
```

### After (on-chain auth)

```solidity
import {SiweAuth} from  "@oasisprotocol/sapphire-contracts/contracts/auth/SiweAuth.sol"

contract MyContract is SiweAuth {
  address private _owner;

  modifier onlyOwner(bytes calldata bearer) {
    if (authMsgSender(bearer) != _owner) {
      revert("not allowed");
    }
    _;
  }

  constructor(string memory domain) SiweAuth(domain) {
    _owner = msg.sender;
  }

  function getSecretMessage(bytes calldata bearer) external view onlyOwner(bearer) returns (string memory) {
    return "Very secret message";
  }
}
```

### DIFF
--- a/contracts/contracts/SiweParser.sol
+++ b/contracts/contracts/SiweParser.sol
@@ -1,0 +1,257 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+import {DateTime} from "./DateTime.sol";
+
+struct ParsedSiweMessage {
+    bytes schemeDomain;
+    address addr;
+    bytes statement;
+    bytes uri;
+    bytes version;
+    uint256 chainId;
+    bytes nonce;
+    bytes issuedAt;
+    bytes expirationTime;
+    bytes notBefore;
+    bytes requestId;
+    bytes[] resources;
+}
+
+/**
+ * @title On-chain parser for EIP-4361 SIWE message
+ * @notice Call parseSiweMsg() and provide the EIP-4361 SIWE message. The parser
+ * will generate the ParsedSiweMessage struct which you can then use to
+ * extract the authentication information in your on-chain contract.
+ */
+library SiweParser {
+    /// Invalid length of the hex-encoded address
+    error InvalidAddressLength();
+    /// Invalid length of the nonce
+    error InvalidNonce();
+
+    /**
+     * @notice Convert string containing hex address without 0x prefix to solidity address object.
+     */
+    function _hexStringToAddress(bytes memory s)
+        internal
+        pure
+        returns (address)
+    {
+        if (s.length != 40) {
+            revert InvalidAddressLength();
+        }
+
+        bytes memory r = new bytes(s.length / 2);
+        for (uint256 i = 0; i < s.length / 2; ++i) {
+            r[i] = bytes1(
+                _fromHexChar(uint8(s[2 * i])) *
+                    16 +
+                    _fromHexChar(uint8(s[2 * i + 1]))
+            );
+        }
+        return address(bytes20(r));
+    }
+
+    function _fromHexChar(uint8 c) internal pure returns (uint8) {
+        if (bytes1(c) >= bytes1("0") && bytes1(c) <= bytes1("9")) {
+            return c - uint8(bytes1("0"));
+        }
+        if (bytes1(c) >= bytes1("a") && bytes1(c) <= bytes1("f")) {
+            return 10 + c - uint8(bytes1("a"));
+        }
+        if (bytes1(c) >= bytes1("A") && bytes1(c) <= bytes1("F")) {
+            return 10 + c - uint8(bytes1("A"));
+        }
+        return 0;
+    }
+
+    /**
+     * @notice Substring.
+     */
+    function _substr(
+        bytes memory str,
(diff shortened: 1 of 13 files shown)


aefhm (commented):
Super clear test cases!

contracts/contracts/SiweParser.sol
```diff
@@ -0,0 +1,252 @@
...
+
+        // dApp Domain.
+        for (; i < siweMsg.length; i++) {
+            if (siweMsg[i] == " ") {
+                p.schemeDomain = _substr(siweMsg, 0, i);
+                break;
+            }
+        }
+
+        i += 50; // " wants you to sign in with your Ethereum account:\n"
```
aefhm, line 193: Wow. They really thought this out.

contracts/test/auth.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  // Signs the given message as ERC-191 "personal_sign" message.
+  async function erc191sign(msg: string, account: Signer) {
+    return ethers.Signature.from(await account.signMessage(msg));
+  }
+
+  it('Should login', async function () {
+    // Skip this test on non-sapphire chains.
+    // It require on-chain encryption and/or signing.
+    if ((await ethers.provider.getNetwork()).chainId == 31337n) {
```
aefhm, line 46: Not whitelist?

contracts/test/auth.ts
```diff
@@ -0,0 +1,169 @@
...
+        await erc191sign(siweStrWrongSig, account),
+      ),
+    ).to.be.reverted;
+
+    // Expired login.
+    let now = new Date();
+    const siweStrExpired = await siweMsg(
+      'localhost',
+      1,
+      new Date(now.getFullYear(), now.getMonth() - 1, 1),
```
aefhm, line 89: We didn't want to edge this more?

contracts/contracts/auth/SiweAuth.sol
```diff
@@ -0,0 +1,171 @@
...
+ *
+ *   modifier onlyOwner(bytes calldata bearer) {
+ *     if (authMsgSender(bearer) != _owner) {
+ *       revert("not allowed");
+ *     }
+ *     _;
+ *   }
+ *
+ *   constructor(string memory domain) SiweAuth(domain) {
+ *     _owner = msg.sender;
```
aefhm, line 36: I wonder if it would be clearer if `onlyOwner` were also extended or duplicated to `onlyManager` as well?

contracts/contracts/SiweParser.sol
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        i++; // New line.
+
+        (p.uri, i) = _parseField(siweMsg, "URI", i);
+        (p.version, i) = _parseField(siweMsg, "Version", i);
+        bytes memory chainId;
+        (chainId, i) = _parseField(siweMsg, "Chain ID", i);
+        p.chainId = _parseUint(chainId);
+        (p.nonce, i) = _parseField(siweMsg, "Nonce", i);
```
aefhm, line 218: > nonce REQUIRED. A random string typically chosen by the relying party and used to prevent replay attacks, at least 8 alphanumeric characters.

Probably too pedantic?

aefhm (approved):
Comments aren't blocking.

contracts/contracts/SiweParser.sol
matevz: Yeah, the standard wants to keep a human-readable message in case the wallet doesn't support any fancy UIs around SIWE.

contracts/contracts/auth/SiweAuth.sol
matevz: That would just add clutter IMO. The example code is the minimal code to demonstrate the usage of SiweAuth.

contracts/test/auth.ts
matevz: Fixed.

PR #341, opened 2024-07-26
contracts: Fix subcall undelegation done

aefhm:
## Description

We passed in the kind of `UndelegateStart` or 1 when we need `UndelegateDone` or 2.

## TODO

- [x] Figure out why tests didn't catch this

### DIFF
--- a/contracts/contracts/Subcall.sol
+++ b/contracts/contracts/Subcall.sol
@@ -398,7 +398,7 @@ library Subcall {
         returns (uint128 amount)
     {
         bytes memory result = consensusTakeReceipt(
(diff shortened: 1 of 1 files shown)


aefhm:
Our tests only covered the decoding. https://github.com/oasisprotocol/sapphire-paratime/blob/ff1b8f99ae7f852a34841b3f54e4238ffae2cc7b/contracts/contracts/tests/SubcallTests.sol#L55-L61

lubej (approved):
Can confirm this working in https://github.com/oasisprotocol/dapp-staker/pull/3

PR #346, opened 2024-08-09
workflow: Harden GitHub workflow against injection

lukaw3d:
## Description

Use environment variable to prevent injection [attack](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable).


kostko (approved):

aefhm (approved):
Thanks Luka!

PR #354, opened 2024-08-21
Check, if contracts-test fail on solidity errors

matevz:
Debug PR, if contracts-test fails, if there are syntax errors in Solidity contracts.


matevz:
Works as expected. Closing.

Issue #351, opened 2024-08-19
docs: Sapphire quickstart guide: vigil.sol and run-vigil.ts links broken 


CdavM:
I have verified that the issue has not been reported yet.

##### SUMMARY

https://github.com/oasisprotocol/sapphire-paratime/blob/main/docs/quickstart.mdx links to vigil.sol and run-vigil.ts, but the links are broken on the main branch.

##### ISSUE TYPE
- Bug Report

CdavM:
Workaround: use an alternative branch like https://github.com/oasisprotocol/sapphire-paratime/blob/CedarMist/wagmi-viem/examples/hardhat/scripts/run-vigil.ts

aefhm:
> Workaround: use an alternative branch like https://github.com/oasisprotocol/sapphire-paratime/blob/CedarMist/wagmi-viem/examples/hardhat/scripts/run-vigil.ts

Agree and thanks for the suggestion! Using this workaround in https://github.com/oasisprotocol/sapphire-paratime/pull/359.

aefhm:
We should also rethink the entire [quick start](https://docs.oasis.io/dapp/sapphire/quickstart) section, but that should be a separate issue/ticket.

aefhm:
Closed by https://github.com/oasisprotocol/sapphire-paratime/pull/378

PR #359, opened 2024-08-23
docs: Update quickstart code references

aefhm:
## Description

After Viem and Wagmi V2 changes in https://github.com/oasisprotocol/sapphire-paratime/pull/303, some code examples are out of date. Closes https://github.com/oasisprotocol/sapphire-paratime/issues/351

### DIFF
--- a/docs/quickstart.mdx
+++ b/docs/quickstart.mdx
@@ -112,8 +112,8 @@ Start by pasting [Vigil.sol] into `contracts/Vigil.sol`.
 While you're there, also place [run-vigil.ts] into `scripts/run-vigil.ts`.
 We'll need that later.
 
-[Vigil.sol]: https://github.com/oasisprotocol/sapphire-paratime/blob/main/examples/hardhat/contracts/Vigil.sol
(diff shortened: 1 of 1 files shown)


matevz:
Instead of using commit hashes, should we make the `clients/js/stable/1.x` branch instead? This is also useful, if we needed to backport anything until 2.x is stable.

aefhm:
> Instead of using commit hashes, should we make the `clients/js/stable/1.x` branch instead? This is also useful, if we needed to backport anything until 2.x is stable.

In which case, I think we would branch off this [commit](https://github.com/oasisprotocol/sapphire-paratime/commit/ff1b8f99ae7f852a34841b3f54e4238ffae2cc7b)? I think `main` is quite a bit ahead by now. Would this slightly more readable [alternative](https://github.com/oasisprotocol/sapphire-paratime/blob/clients/js/v1.3.2/examples/hardhat/contracts/Vigil.sol) (than just a commit SHA) be reasonable?

aefhm:
Superceded by https://github.com/oasisprotocol/sapphire-paratime/pull/378

PR #360, opened 2024-08-26
contracts: silence SIWE Solidity tests compile warnings

CedarMist:
* Updates `clients/py` requirements to specify Web3.py v6.x
 * Fixes warnings in `contracts/tests/SiweParserTests.sol`

### DIFF
--- a/clients/py/requirements.txt
+++ b/clients/py/requirements.txt
@@ -1,3 +1,3 @@
 cbor2
 pynacl
-web3
+web3==6.*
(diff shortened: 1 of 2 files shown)


CedarMist:
Apparently the Python client isn't compatible with Web3.py 7.x and we're not pinning the version in `requirements.txt`

7.x was released Aug 21, prior to that 6.x was being used

matevz (approved):
LGTM.
cc @ZigaMr

Issue #362, opened 2024-08-26
Fix packaging releases, use `pnpm` rather than `npm`


CedarMist:
The following packages aren't released properly:

 * https://bundlephobia.com/package/@oasisprotocol/sapphire-ethers-v6
 * https://bundlephobia.com/package/@oasisprotocol/sapphire-wagmi-v2
 * https://bundlephobia.com/package/@oasisprotocol/sapphire-viem-v2

Their dependencies in `package.json` show `workspace:^` because they're released with NPM rather than PNPM.

In https://github.com/oasisprotocol/sapphire-paratime/blob/main/.github/workflows/publish.yaml it should use `pnpm` to publish the packages.

We should also enable provenance when publishing packages.

PR #358, opened 2024-08-23
workflow: Fix Ethers V6 publish regex

aefhm:
## Description

We needed to add the `ethers-v6` JS package to the publishing regex.

PR #363, opened 2024-08-26
packaging: use PNPM for publishing, with correct tags

CedarMist:
fixes: #362
includes: #358

extracts the following tags:

 * rc
 * next
 * alpha
 * beta

also enables NPM provenance when publishing

We can change the dist-tag for the `latest` tag: https://stackoverflow.com/questions/67813980/how-to-point-latest-dist-tag-to-a-previous-version-in-npm

```bash
npm dist-tag add <pkg name>@<version> latest
```

And release new versions under the appropriate `alpha` or `beta` tags. Maybe as @matevz recommended we should just use `next` ?

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -47,12 +47,11 @@ jobs:
           # https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable
           REF_NAME: ${{ github.ref_name }}
         run: |
-          echo "NPM_PACKAGE=$(echo $REF_NAME | grep -oE '(clients/js|contracts|integrations/(hardhat|wagmi-v2|viem-v2))')" >> $GITHUB_OUTPUT
+          echo "NPM_PACKAGE=$(echo $REF_NAME | grep -oE '(clients/js|contracts|integrations/(ethers-v6|hardhat|wagmi-v2|viem-v2))')" >> $GITHUB_OUTPUT
+          echo "NPM_TAG=$(echo $REF_NAME | grep -oP '(?<=\-)(rc|next|alpha|beta)' || echo 'latest')" >> $GITHUB_OUTPUT
       - name: Publish ${{ github.ref_name }} to NPM
-        uses: JS-DevTools/npm-publish@v3
-        with:
-          # Build runs as a prepublish script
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -47,12 +47,11 @@ jobs:
...
+          echo "NPM_TAG=$(echo $REF_NAME | grep -oP '(?<=\-)(rc|next|alpha|beta)' || echo 'latest')" >> $GITHUB_OUTPUT
       - name: Publish ${{ github.ref_name }} to NPM
-        uses: JS-DevTools/npm-publish@v3
-        with:
-          # Build runs as a prepublish script
-          ignore-scripts: false
-          token: ${{ secrets.NPM_TOKEN }}
-          package: ${{ steps.extract-tag.outputs.NPM_PACKAGE }}
-          access: public
+        run: pnpm publish --access public --no-git-checks --tag ${{ steps.extract-tag.outputs.NPM_TAG }}
```
lukaw3d, line 53: steps.extract-tag.outputs.NPM_TAG can be unsafe; move into env

Related https://github.com/oasisprotocol/sapphire-paratime/pull/346

CedarMist: $REF_NAME will only ever be one of a number of known values, thus it is a trusted input.

CedarMist: Additionally, which tags can be created are restricted, so even `github.ref_name` is a trusted input.

Originally the regexes were much stricter. I can change the pattern matchers, e.g. 

```
      - integrations/wagmi-v2/v[0-9]+.[0-9]+.[0-9]+
      - integrations/wagmi-v2/v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+
      - integrations/wagmi-v2/v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+
      - integrations/wagmi-v2/v[0-9]+.[0-9]+.[0-9]+-next.[0-9]+
```

etc.

aefhm (approved):
pnpm usage makes sense to me. I'm open to either resolution with regards to `REF_NAME` usage.

.github/workflows/publish.yaml
CedarMist: Resolving as have moved it to environment variable

PR #365, opened 2024-08-27
packaging: bump versions + add CHANGELOG entries

CedarMist:
re: #362
re: #363

for the following packages:

 * oasisprotocol/sapphire-paratime
 * oasisprotocol/sapphire-ethers-v6
 * oasisprotocol/sapphire-hardhat
 * oasisprotocol/sapphire-viem-v2
 * oasisprotocol/sapphire-wagmi-v2

They will be released under the `next` tag, to test updated publish.yaml workflow
after publishing the bundlephobia links in the main README.md should work
It should also be possible to install the packages & resolve their dependencies.

### DIFF
--- a/integrations/hardhat/CHANGELOG.md
+++ b/integrations/hardhat/CHANGELOG.md
@@ -1,0 +1,13 @@
+# Change Log
+
+All notables changes to this project are documented in this file.
+
+The format is inspired by [Keep a Changelog].
+
+[Keep a Changelog]: https://keepachangelog.com/en/1.0.0/
+
+## 2.22.2-next.0 (2024-08)
+
+### Fixed
+
+- Use PNPM to publish packaging, so `workspace:^` links in package.json get translated to their correct versions.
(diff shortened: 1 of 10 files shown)


integrations/hardhat/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Change Log
+
+All notables changes to this project are documented in this file.
+
+The format is inspired by [Keep a Changelog].
+
+[Keep a Changelog]: https://keepachangelog.com/en/1.0.0/
+
+## 2.22.2-next.0 (2024-08)
```
aefhm, line 9: I guess `next` is a direct replacement for `alpha` then.

aefhm (approved):
I suppose we could have also followed NPM thinking and just tagged `next` release on NPM and not used SemVer pre-release language. 🤔

integrations/hardhat/CHANGELOG.md
CedarMist: Sorry I had to investigate this a bit, so we don't need to specify this in package.json.

We can just rely on tags, and keep the version in package.json as e.g. `v2.22.2`

CedarMist: I'm going to stick with this for now, as it's part of the semver spec and has specific meaning for prereleases

https://www.npmjs.com/package/semver#user-content-prerelease-tags

PR #370, opened 2024-08-29
contracts: add support for core.CallDataPublicKey and core.CurrentEpoch

CedarMist:
Adds two methods to `Subcall.sol`

 * `coreCallDataPublicKey`
 * `coreCurrentEpoch`

### DIFF
--- a/contracts/test/subcall.ts
+++ b/contracts/test/subcall.ts
@@ -5,6 +5,7 @@ import * as cborg from 'cborg';
 import { SubcallTests } from '../typechain-types/contracts/tests/SubcallTests';
 import { SignerWithAddress } from '@nomicfoundation/hardhat-ethers/signers';
 import {
+  AbiCoder,
   ContractTransactionReceipt,
   EventLog,
   Provider,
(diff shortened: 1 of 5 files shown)


lubej (approved):

contracts/test/subcall.ts
```diff
@@ -491,4 +495,73 @@ describe('Subcall', () => {
...
+    expect(subcall_publickey.key).lengthOf(32);
+    expect(subcall_publickey).has.haveOwnProperty('checksum');
+    expect(subcall_publickey.checksum).lengthOf(32);
+    expect(subcall_publickey).has.haveOwnProperty('signature');
+    expect(subcall_publickey.signature).lengthOf(64);
+    expect(subcall_publickey).has.haveOwnProperty('expiration');
+
+    // Verify CBOR parsing via contract returns the same result
+    const result = await contract.testCoreCallDataPublicKey.staticCall();
+    expect(result.epoch).eq(subcall_data.epoch);
```
matevz, line 548: Did you encounter any issues for epoch changing in the middle of running the test? To mitigate it maybe add a helper like this at the top?

```ts
let result, subcall_data;
do {
  result = await contract.testCoreCallDataPublicKey.staticCall();
  [_, subcall_data] = await provider.call({...}).decode();
} while (result.epoch != subcall_data.epoch);

// Then do actual tests.
```

matevz (approved):
Good job!

For production networks we should add a Web3 client monitoring Testnet and Mainnet call data public keys here: https://github.com/oasisprotocol/network-monitoring/tree/main/web3

contracts/test/subcall.ts
CedarMist: There is such a small amount of time between the two calls and neither of them are transactions, so I think it should be fine.

Likewise with the current epoch test, where I query the docker container directly.

If it turns into an intermittent problem we can make a fix for it easily.

CedarMist:
> Good job!
> 
> For production networks we should add a Web3 client monitoring Testnet and Mainnet call data public keys here: https://github.com/oasisprotocol/network-monitoring/tree/main/web3

I'll be using the subcall (directly, not via the contract) in the go, python & js clients, so if encrypted transactions are being performed it will have validated the calldata public key and successfully used it.

PR #372, opened 2024-08-30
use `core.CallDataPublicKey` subcall, not `oasis_callDataPublicKey` RPC call

### LINKED ISSUE #367: Retrieve calldata public key using subcall rather than `oasis_callDataPublicKey` RPC call
This will prevent warnings in the console when MetaMask rejects custom RPC calls.

This will also removed code that detects the chain and uses a JSON-RPC client to manually perform the `oasis_callDataPublicKey` RPC call.

This will support other runtimes without needing to modify the client code.

It will also support highly firewalled environments, non-standard RPC nodes etc. more easily as the fallback code has been removed which performs the `oasis_callDataPublicKey` request. 

---

This is only really necessary for the JS client, as the Python & Go clients don't have MetaMask denying specific RPC calls.

CedarMist:
Fixes #367


CedarMist:
So, we have a build bug with bigint literals:

```
Error: src/calldatapublickey.ts(59,46): error TS2737: BigInt literals are not available when targeting lower than ES2020.
```

e.g.

```typescript
return bytes.reduce((acc, byte) => (acc << 8n) | BigInt(byte), 0n);
```

Not sure if we should change the target, or remove the bigint literals. I'm surprised we haven't seen this bug before.
The place it's being used is for manual parsing of uint256 from abi results (so as to not pull in a full ABI decoding library). The expected values are going to be small, and certainly within a Javascript Number's safe-int-range

clients/js/src/provider.ts
```diff
@@ -168,7 +169,16 @@ export function makeSapphireRequestFn(
 
     // Decrypt responses which return encrypted data
     if (method === 'eth_call') {
-      return cipher.decryptResult(res as BytesLike);
+      // If it's an unencrypted core.CallDataPublicKey query, don't attempt to decrypt the response
+      const isCDPKQuery =
```
aefhm, line 173: Helper function?

aefhm:
> So, we have a build bug with bigint literals:
> 
> ```
> Error: src/calldatapublickey.ts(59,46): error TS2737: BigInt literals are not available when targeting lower than ES2020.
> ```
> 
> e.g.
> 
> ```ts
> return bytes.reduce((acc, byte) => (acc << 8n) | BigInt(byte), 0n);
> ```
> 
> Not sure if we should change the target, or remove the bigint literals. I'm surprised we haven't seen this bug before. The place it's being used is for manual parsing of uint256 from abi results (so as to not pull in a full ABI decoding library). The expected values are going to be small, and certainly within a Javascript Number's safe-int-range

I think I'm comfortable going up.

clients/js/test/utils.ts
```diff
@@ -28,20 +28,29 @@ export class MockEIP1193Provider {
   public constructor(public chainId: number) {
     this.calldatakeypair = nacl.sign.keyPair();
     this.request = jest.fn(async ({ method, params }) => {
-      if (method === OASIS_CALL_DATA_PUBLIC_KEY) {
+      // Intercept calls to the `core.CallDataPublicKey` subcall
+      if (
```
aefhm, line 32: We don't use the `isCallDataPublicKeyQuery` function but test implicitly?

aefhm (approved):

Issue #361, opened 2024-08-26
Intermittent failure in contracts/test/auth.ts


CedarMist:
```
  1) Auth
       Should call authenticated method:
     TypeError: Expected a valid transaction hash, but got 'Very secret message'
      at onSuccess (/home/runner/work/sapphire-paratime/sapphire-paratime/node_modules/.pnpm/@nomicfoundation+hardhat-chai-matchers@2.0.2_@nomicfoundation+hardhat-ethers@3.0.5_chai@4.3.7_djiantmfwzd4h5mvs2d3iacurm/node_modules/@nomicfoundation/hardhat-chai-matchers/src/internal/reverted/reverted.ts:32:17)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (test/auth.ts:166:5)
```

Unable to reproduce locally.

https://github.com/oasisprotocol/sapphire-paratime/blob/cf0cdd51df9c3c3ec8e6fb080a77eb057faa34e4/contracts/test/auth.ts#L151-L166

Presumably waiting 2 blocks isn't enough, it should sleep until the timestamp has really passed, and the bearer token will have expired.

@matevz

PR #375, opened 2024-09-03
contracts: fix intermittent failures in auth.ts tests

CedarMist:
fixes: #361

Also fixes typescript warnings in auth.ts

### DIFF
--- a/contracts/test/auth.ts
+++ b/contracts/test/auth.ts
@@ -6,18 +6,20 @@ import { Signer } from 'ethers';
 import { SiweMessage } from 'siwe';
 import '@nomicfoundation/hardhat-chai-matchers';
 
-import { SiweAuthTests__factory } from '../typechain-types/factories/contracts/tests';
-import { SiweAuthTests } from '../typechain-types/contracts/tests/auth/SiweAuthTests';
 import { NETWORKS } from '@oasisprotocol/sapphire-paratime';
+import { HardhatNetworkHDAccountsConfig } from 'hardhat/types';
+import { SiweAuthTests } from '../typechain-types/contracts/tests/auth/SiweAuthTests';
+
+function delay(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
 
 describe('Auth', function () {
   async function deploy(domain: string) {
-    const SiweAuthTests_factory = await ethers.getContractFactory(
-      'SiweAuthTests',
-    );
-    const siweAuthTests = await SiweAuthTests_factory.deploy(domain);
+    const factory = await ethers.getContractFactory('SiweAuthTests');
+    const siweAuthTests = await factory.deploy(domain);
     await siweAuthTests.waitForDeployment();
-    return siweAuthTests;
+    return siweAuthTests as unknown as SiweAuthTests;
   }
 
   async function siweMsg(
@@ -54,7 +56,8 @@ describe('Auth', function () {
     const siweAuthTests = await deploy('localhost');
 
     // Correct login.
-    const accounts = config.networks.hardhat.accounts;
+    const accounts = config.networks.hardhat
+      .accounts as HardhatNetworkHDAccountsConfig;
     const account = ethers.HDNodeWallet.fromMnemonic(
       ethers.Mnemonic.fromPhrase(accounts.mnemonic),
       accounts.path + '/0',
@@ -113,7 +116,8 @@ describe('Auth', function () {
     const siweAuthTests = await deploy('localhost');
 
     // Author should read a very secret message.
-    const accounts = config.networks.hardhat.accounts;
+    const accounts = config.networks.hardhat
+      .accounts as HardhatNetworkHDAccountsConfig;
     const account = ethers.HDNodeWallet.fromMnemonic(
       ethers.Mnemonic.fromPhrase(accounts.mnemonic),
       accounts.path + '/0',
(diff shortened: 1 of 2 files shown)


contracts/test/auth.ts
```diff
@@ -149,20 +154,14 @@ describe('Auth', function () {
...
-    // Wait for 2 blocks. The first block may still have a timestamp earlier than the expiration time.
-    await new Promise(async (r) => {
-      const desiredBlockNumber = (await ethers.provider.getBlockNumber()) + 2;
-      ethers.provider.on('block', (blockNumber) => {
-        if (blockNumber == desiredBlockNumber) {
-          r();
-        }
-      });
-    });
+    await delay(expiration.getTime() - new Date(Date.now()).getTime());
```
lukaw3d, line 163: `new Date(Date.now()).getTime()` looks like a typo but works. It's equal to `new Date().getTime()` and `Date.now()` tho

lukaw3d (approved):

contracts/test/auth.ts
CedarMist: Yup it's equivalent. I think I had a minor brainfart.

contracts/test/auth.ts
```diff
@@ -149,20 +154,14 @@ describe('Auth', function () {
...
-    await new Promise(async (r) => {
-      const desiredBlockNumber = (await ethers.provider.getBlockNumber()) + 2;
-      ethers.provider.on('block', (blockNumber) => {
-        if (blockNumber == desiredBlockNumber) {
-          r();
-        }
-      });
-    });
+    await delay(expiration.getTime() - Date.now());
+    await siweAuthTests.doNothing();
```
aefhm, line 164: We need this call? 🤔

aefhm (approved):

contracts/test/auth.ts
CedarMist: Block times aren't metronomic and the time experienced in view calls is that of whatever the last mined block was.

So I'm waiting until I 100% know that it will be expired, then I push a transaction to force a block to be mined and to wait for the next block, then the view call executes in the context of the last block whos timestamp will always be after the bearer token expiration.

CedarMist:
Well... F... 

This is still running into the problem! We're waiting for bearer token to expire, push a transaction through... And it still thinks the bearer token is valid?

```
  1) Auth
       Should call authenticated method:
     TypeError: Expected a valid transaction hash, but got 'Very secret message'
      at onSuccess (/home/runner/work/sapphire-paratime/sapphire-paratime/node_modules/.pnpm/@nomicfoundation+hardhat-chai-matchers@2.0.2_@nomicfoundation+hardhat-ethers@3.0.5_chai@4.3.7_djiantmfwzd4h5mvs2d3iacurm/node_modules/@nomicfoundation/hardhat-chai-matchers/src/internal/reverted/reverted.ts:32:17)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (test/auth.ts:165:5)
```

CedarMist:
@matevz Any idea why this test is still failing intermittently?

lukaw3d:
> This is still running into the problem

I tried repeating the test
https://github.com/oasisprotocol/sapphire-paratime/actions/runs/10692954396/job/29642295984
failed 6/10 repeats

I tried double `await siweAuthTests.doNothing();`
https://github.com/oasisprotocol/sapphire-paratime/actions/runs/10693129388/job/29642755780
failed 1/10 repeats

I tried `await delay(6000);`
https://github.com/oasisprotocol/sapphire-paratime/actions/runs/10694436081/job/29646226736
it succeeded 20/20 repeats

lukaw3d:
https://github.com/oasisprotocol/sapphire-paratime/actions/runs/10692954396/job/29642295984
also randomly failed https://github.com/oasisprotocol/sapphire-paratime/blob/58be8a6ad66f9532ddb3104a95c369e7724534dd/contracts/test/subcall.ts#L318-L320
(I tried repeating it 20 times but it just started failing differently after 10th success :shrug: https://github.com/oasisprotocol/sapphire-paratime/actions/runs/10694614566/job/29646662356 )

contracts/test/auth.ts
```diff
@@ -123,7 +128,7 @@ describe('Auth', function () {
       siweStr,
       await erc191sign(siweStr, account),
     );
-    await expect(await siweAuthTests.testVerySecretMessage(bearer)).to.be.equal(
+    expect(await siweAuthTests.testVerySecretMessage(bearer)).to.be.equal(
```
matevz, line 131: Is there a difference between to.be.equal and to.be.reverted for sync/async behavior?

matevz (approved):
Nice catch, thanks.

contracts/test/auth.ts
matevz: I think ` await siweAuthTests.doNothing();` doesn't assure you that the block will be mined, just that the tx is submitted? You need to wait for the receipt.

matevz:
> @matevz Any idea why this test is still failing intermittently?

https://github.com/oasisprotocol/sapphire-paratime/pull/375#discussion_r1743400961

contracts/test/auth.ts
CedarMist: Yes, reverted takes a promise, you await the `expect`.
Otherwise, you await the result, then `expect` it.

contracts/test/auth.ts
CedarMist: Good catch, somehow that slipped my mind.

CedarMist:
So, even if I wait for the tx receipt from `doNothing`, I still get intermittent failures.

CedarMist:
So to summarize:

 * Wasn't waiting for tx to be mined by waiting for receipt
 * Was providing milliseconds, which rounded to seconds
 * On-chain time is in seconds
 * Wasn't waiting an extra second, to account for possible round-up

Have now run this 20 times in a row with no failures.

CedarMist:
I'll add in more debugging to check:

 * timestamp in SIWE message
 * timestamp, block number & epoch returned from view call
 * timestamp of block etc.

PR #379, opened 2024-09-04
integrations: Remove extra fetching of runtime public key

aefhm:
## Description

Address Node timeout issue encountered by [developers](https://discord.com/channels/748635004384313474/960599114993762304/1279839606719643650).

> Im using the alpha of @oasisprotocol/sapphire-viem-v2, and it looks like when using it, node process wont naturally terminate (w/o a call to exit() ). I think that its the fault of this timer

https://github.com/oasisprotocol/sapphire-paratime/blob/77a2bea6b11e8cbd87da8c15c725ee01b77b381b/integrations/viem-v2/src/index.ts#L126-L128

## Resolution

I don't think we need the second `fetch`.

### DIFF
--- a/integrations/viem-v2/src/index.ts
+++ b/integrations/viem-v2/src/index.ts
@@ -123,9 +123,6 @@ export async function createSapphireSerializer<
 	const fetcher = new KeyFetcher();
 	const provider = client as EthereumProvider;
 	await fetcher.fetch(provider);
-	setTimeout(async () => {
-		await fetcher.fetch(provider);
-	}, fetcher.timeoutMilliseconds);
 
 	const wrappedSerializer = ((tx, sig?) => {
 		if (!sig) {


CedarMist:
So, this is a workaround for the fact that Viem's code for packing transactions isn't async, meaning when we encrypt a transaction we can't do an RPC request to retrieve the calldata public key.

We do the first fetch to retrieve the public key, and then the timer proactively fetches public keys.

Without the timer and periodic public key fetching the app will eventually start encrypting transactions with expired keys causing it to fail.

It would be good to find a workaround for this, as it's not the best. But just removing it will cause long-running dApps to break.

CedarMist:
It would be good to setup a test case for using Viem from Node CLI to reproduce.

And use https://www.npmjs.com/package/wtfnode to see which handles are still active.

Could use a `WeakRef` to allow the callback function to be garbage collected, but that's not 100% predictable, see:

```typescript
let callback = () => {
  console.log('Interval tick');
  // Your interval logic here
};

const weakCallback = new WeakRef(callback);

const intervalId = setInterval(() => {
  const cb = weakCallback.deref();
  if (cb) {
    cb();
  } else {
    console.log('Callback has been garbage collected');
    clearInterval(intervalId);
  }
}, 1000);

// At some point later
callback = null; // Allow the callback to be garbage collected
```

Or we could trap `SIGINT` via `process.on` to clear the interval... Would need to detect if it's Node or Web via checking if `globalThis.process` is undefined.

CedarMist:
Or... we could refactor the Viem code to not do any transaction encryption inside the non-async Viem code, and push it all into the provider code? However, last time I tried that it had trouble as it wasn't always going through the provider and I had trouble intercepting calls to the provider.

integrations/viem-v2/src/index.ts
```diff
@@ -123,9 +123,6 @@ export async function createSapphireSerializer<
 	const fetcher = new KeyFetcher();
 	const provider = client as EthereumProvider;
 	await fetcher.fetch(provider);
-	setTimeout(async () => {
```
CedarMist, line 126: Lets document the reason for having the interval timer.

It be changed to `setInterval`, so it's continuous.

And the interval ID should be `unref`'d

CedarMist (changes requested):

integrations/viem-v2/src/index.ts
aefhm: Agree. `unref` seems best right now.

CedarMist: Have you been able to reproduce the problem with a test? Would be good to have this in CI

matevz:
Closing this in favor of https://github.com/oasisprotocol/sapphire-paratime/pull/383

PR #383, opened 2024-09-06
clients/viem: use interval for routine fetching, unref intervalId

CedarMist:
Fixes #379

### DIFF
--- a/integrations/viem-v2/src/index.ts
+++ b/integrations/viem-v2/src/index.ts
@@ -119,13 +119,20 @@ export async function createSapphireSerializer<
 		return originalSerializer;
 	}
 
-	// As the serialized is synchronous, fetching keys while running
+	// As the serializer is synchronous, fetching keys while running
 	const fetcher = new KeyFetcher();
 	const provider = client as EthereumProvider;
 	await fetcher.fetch(provider);
-	setTimeout(async () => {
+
+	// The fetcher runs in the background, routinely fetching the keys
+	// This means when the serializer requests a calldata public key one will
+	// have been retrieved pre-emptively.
+	const intervalId = setInterval(async () => {
 		await fetcher.fetch(provider);
 	}, fetcher.timeoutMilliseconds);
+	// The interval ID is unreferenced to prevent Node from hanging at exit
+	// See discussion on https://github.com/oasisprotocol/sapphire-paratime/pull/379
+	intervalId.unref();
 
 	const wrappedSerializer = ((tx, sig?) => {
 		if (!sig) {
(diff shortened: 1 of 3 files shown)


CedarMist:
This won't work in browsers.

We can do something like:

```typescript
if( typeof x === 'object' && 'unref' in x && typeof x.unref == 'function' )
```

or just
```typescript
if( typeof x.unref === 'function' )
```

?

CedarMist:
Will find a way of linting the TS code for NodeJS vs WebBrowser compatibility.

CedarMist:
Have logged a ticket/discussion with Viem w.r.t async serializers, which would entirely avoid this hacky workaround. I should've done this earlier.

https://github.com/wevm/viem/discussions/2696

integrations/viem-v2/src/index.ts
```diff
@@ -119,13 +119,23 @@ export async function createSapphireSerializer<
...
+	// The fetcher runs in the background, routinely fetching the keys
+	// This means when the serializer requests a calldata public key one will
+	// have been retrieved pre-emptively.
+	const intervalId: NodeJS.Timeout | number = setInterval(async () => {
 		await fetcher.fetch(provider);
 	}, fetcher.timeoutMilliseconds);
+	// The interval ID is unreferenced to prevent Node from hanging at exit
+	// See discussion on https://github.com/oasisprotocol/sapphire-paratime/pull/379
+	// This is only available in NodeJS, and not in browsers
+	if (typeof intervalId.unref === "function") {
```
aefhm, line 136: I like the upstream fix best. I think this patch's compromise is reasonable.

aefhm:
A standalone NodeJS script reproduces the exit error pretty easily, but reproducing it in `vitest` has been tricky. Test harness is not waiting for the Node process, and does not hang which is the behavior we want to reproduce in the failing case. We could use a test script and assert that this exits within 1000 milliseconds though.

```TypeScript
const transport = sapphireHttpTransport();
const chain = sapphireLocalnet;
const publicClient = createPublicClient({ chain, transport });
defineChain({
    id: 0x5afd,
    name: "Oasis Sapphire Localnet",
    network: "sapphire-localnet",
    nativeCurrency: { name: "Sapphire Local Rose", symbol: "TEST", decimals: 18 },
    rpcUrls: {
        default: {
            http: ["http://localhost:8545"],
        },
    },
    serializers: {
        transaction: createSapphireSerializer(publicClient)
    },
    testnet: true,
});
```

CedarMist:
Ok, I can reproduce this.

aefhm (approved):

PR #396, opened 2024-09-17
Add signed data/leash/signature to eth_call data

ZigaMr:
This PR adds signed queries to sapphire-py wrapper. Based on Go client, it builds a data_pack dictionary containing the signature and eip712 message data. 
Currently returns "execution_reverted" error when testing on sapphire-localnet.


clients/py/sapphirepy/sapphire.py
```diff
@@ -63,79 +87,152 @@ def _encrypt_tx_params(pk:CalldataPublicKey, params:tuple[TxParams]):
...
+        msg_data = {
+            "from": params[0].get('from'),
+            "to": params[0].get('to', '0x'),
+            "value": params[0].get('value', 0),
+            "gasLimit": params[0].get('gas', DEFAULT_GAS_LIMIT),
+            "gasPrice": params[0].get('gasPrice', DEFAULT_GAS_PRICE),
+            "data": params[0].get('data'),
+            "leash":
+                {
+                    "nonce": web3.eth.get_transaction_count(params[0].get('from')),
```
CedarMist, line 126: Extract block and nonce into variable, to prevent multiple RPC requests for same data.

clients/py/sapphirepy/sapphire.py
```diff
@@ -63,79 +87,152 @@ def _encrypt_tx_params(pk:CalldataPublicKey, params:tuple[TxParams]):
...
+                        # If no calldata public key exists, fetch one
+                        cdpk = cast(RPCResponse, make_request(RPCEndpoint('oasis_callDataPublicKey'), []))
+                        pk = cast(Optional[CalldataPublicKey], cdpk.get('result', None))
+                        if pk:
+                            manager.add(pk)
+                    if not pk:
+                        raise RuntimeError('Could not retrieve callDataPublicKey!')
+                    do_fetch = False
+
+                    c = _encrypt_tx_params(pk, params, w3,  account)
```
CedarMist, line 200: This makes no distinction between transactions being submitted and `eth_call` . This will break transactions - need two code paths, one for transactions and one for `eth_call`.

clients/py/sapphirepy/sapphire.py
```diff
@@ -63,79 +87,152 @@ def _encrypt_tx_params(pk:CalldataPublicKey, params:tuple[TxParams]):
...
+            'BlockHash': web3.eth.get_block(web3.eth.block_number)['hash'],
+            'BlockRange': DEFAULT_BLOCK_RANGE,
+        }
+        data_pack = {
+            'data': params[0]['data'],
+            # 'data': encrypted_data,
+            'leash': leash,
+            'signature': signed_msg['signature'],
+        }
+        params[0]['data'] = cbor2.dumps(data_pack, canonical=True)
```
CedarMist, line 147: Data isn't encrypted.

CedarMist (changes requested):
Tests are failing

johnny-emp:
i think the issue is that the domain is signed using camelCase but submitted using snake_case

CedarMist:
> i think the issue is that the domain is signed using camelCase but submitted using snake_case

Ah, out-of-band I gave @ZigaMr some code which replicates one of the Go test cases, but in Python:

https://gist.github.com/CedarMist/5319565c32824e56afee5bbbfeeae015

ZigaMr:
> i think the issue is that the domain is signed using camelCase but submitted using snake_case

This is true for the signature part, but I'm still working out the encryption/serialization after the signature is calculated. I tested in real time using fixed PK/ephemeral_pubkey and managed to encrypt the data correctly. I just have to update the data types on last step at cbor2.dump(packedCall) since some bytes aren't matching with the Go (correct) version.

johnny-emp:
i got it working, will share my code in a few

johnny-emp:
still pretty rough, but it has the general idea in here and it's working in my canned test.  I need to find a better way to sign the transactions inside the middleware.  It might be better as an isolated tool for this reason.

https://github.com/empyrealapp/eth-packages/pull/55/commits/12a3c5cb5f0583ab88cb3acebc917e2efa598640

ZigaMr:
> still pretty rough, but it has the general idea in here and it's working in my canned test. I need to find a better way to sign the transactions inside the middleware. It might be better as an isolated tool for this reason.
> 
> [empyrealapp/eth-packages@12a3c5c](https://github.com/empyrealapp/eth-packages/commit/12a3c5cb5f0583ab88cb3acebc917e2efa598640)

I think we both had the same idea :). I misread your msg from earlier, you were correct about the snake_case for submit notation.

CedarMist:
```diff
diff --git a/.github/workflows/contracts-test.yaml b/.github/workflows/contracts-test.yaml
index 8676c28..be77f7b 100644
--- a/.github/workflows/contracts-test.yaml
+++ b/.github/workflows/contracts-test.yaml
@@ -32,6 +32,7 @@ jobs:
         working-directory: clients/py
         run: |
           python3 -mpip install --user -r requirements.txt
+          python3 -mpip install --user -r requirements.dev.txt
           python3 -munittest discover
       - name: Install Node.js
         uses: actions/setup-node@v4
```

matevz (commented):
Please also add sapphirepy compilation and tests to CI. e.g. inside ci-test.yml.

CedarMist:
I'm confused, why do we have two PRs, #433 and #396 ?

clients/py/sapphirepy/sapphire.py
```diff
@@ -69,79 +93,187 @@ def _encrypt_tx_params(pk:CalldataPublicKey, params:tuple[TxParams]):
     else:
         raise TypeError("Invalid 'data' type", type(data))
     encrypted_data = c.encrypt(data_bytes)
-    params[0]['data'] = HexStr('0x' + hexlify(encrypted_data).decode('ascii'))
+
+    if params[0]['from'] and method == 'eth_call':
+        assert account is not None, 'Account must be provided for encrypted calls!'
```
CedarMist, line 98: If no `account` is provided we can do regular `eth_call` with non-signed but encrypted calldata.

If `account` is provided, we can sign the call.

clients/py/sapphirepy/sapphire.py
```diff
@@ -41,23 +59,29 @@ def newest(self):
...
                 return False
     return method in ('eth_estimateGas', 'eth_sendTransaction', 'eth_call')
 
-def _encrypt_tx_params(pk:CalldataPublicKey, params:tuple[TxParams]):
+
+def _encrypt_tx_params(pk: CalldataPublicKey,
+                       params: tuple[TxParams],
+                       method,
+                       web3: Web3,
+                       account: Optional[LocalAccount]) -> TransactionCipher:
```
CedarMist, line 84: Should provide `None` as default argument.

```python
account: Optional[LocalAccount]=None
```

CedarMist (changes requested):

clients/py/sapphirepy/sapphire.py
ZigaMr: Added account check to if statement, now signs data only when account is provided.

clients/py/sapphirepy/sapphire.py
ZigaMr: Added `None` default argument.

clients/py/sapphirepy/tests/test_e2e.py
```diff
@@ -20,9 +20,12 @@
 def compiled_test_contract():
     if not os.path.exists(GREETER_ABI):
         # pylint: disable=import-outside-toplevel
-        from solcx import compile_source    # type: ignore
+        import solcx   # type: ignore
```
CedarMist, line 23: Where is this dependency installed?

CedarMist (approved):

Issue #395, opened 2024-09-16
vite packing of sapphire-ethers-v6 fails


CedarMist:
Discovered this while migrating dapp-sidedao to the 2.x client

```
error during build:
../node_modules/.pnpm/@oasisprotocol+sapphire-ethers-v6@6.0.0_ethers@6.10.0_typescript@5.3.3/node_modules/@oasisprotocol/sapphire-ethers-v6/dist/_esm/index.js (3:9): "fillOptions" is not exported by "../node_modules/.pnpm/@oasisprotocol+sapphire-paratime@1.3.2/node_modules/@oasisprotocol/sapphire-paratime/lib/esm/index.js", imported by "../node_modules/.pnpm/@oasisprotocol+sapphire-ethers-v6@6.0.0_ethers@6.10.0_typescript@5.3.3/node_modules/@oasisprotocol/sapphire-ethers-v6/dist/_esm/index.js".
file: /src/dapp-sidedao/node_modules/.pnpm/@oasisprotocol+sapphire-ethers-v6@6.0.0_ethers@6.10.0_typescript@5.3.3/node_modules/@oasisprotocol/sapphire-ethers-v6/dist/_esm/index.js:3:9
```

 * `@oasisprotocol/sapphire-ethers-v6/dist/_esm/index.js (3:9):`
   * "fillOptions" is not exported by
      * `"../node_modules/.pnpm/@oasisprotocol+sapphire-paratime@1.3.2/node_modules/@oasisprotocol/sapphire-paratime/lib/esm/index.js"`,
         * imported by `"../node_modules/.pnpm/@oasisprotocol+sapphire-ethers-v6@6.0.0_ethers@6.10.0_typescript@5.3.3/node_modules/@oasisprotocol/sapphire-ethers-v6/dist/_esm/index.js"`.

file: /src/dapp-sidedao/node_modules/.pnpm/@oasisprotocol+sapphire-ethers-v6@6.0.0_ethers@6.10.0_typescript@5.3.3/node_modules/@oasisprotocol/sapphire-ethers-v6/dist/_esm/index.js:3:9

1: // SPDX-License-Identifier: Apache-2.0
2: import { AbstractSigner, hexlify } from "ethers";
3: import { fillOptions, isCalldataEnveloped, makeTaggedProxyObject, } from "@oasisprotocol/sapphire-paratime";

CedarMist:
Adding this as a dependency seems to fix:

[...]

PR #401, opened 2024-09-17
integrations/ethers-v6: make sapphire-paratime a non-optional dependency

CedarMist:
Fixes: #395

### DIFF
--- a/integrations/ethers-v6/package.json
+++ b/integrations/ethers-v6/package.json
@@ -48,9 +48,6 @@
 	"peerDependenciesMeta": {
 		"typescript": {
 			"optional": true
-		},
-		"@oasisprotocol/sapphire-paratime": {
-			"optional": true
 		}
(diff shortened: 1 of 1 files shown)


aefhm (approved):
I suppose we always want dev to access `sapphire-paratime` types etc.

CedarMist:
> I suppose we always want dev to access `sapphire-paratime` types etc.

The problem was when I have a frontend project that has only `@oasisprotocol/sapphire-ethers-v6`, the appropriate `@oasisprotocol/sapphire-paratime` dependency isn't installed, meaning when Rollup bundles the frontend it doesn't find the dependencies, because they weren't installed by npm

aefhm:
> meaning when Rollup bundles the frontend it doesn't find the dependencies, because they weren't installed by npm

Sounds weird, but doesn't change anything.

PR #400, opened 2024-09-17
docs: adde networks page

rube-de:
- move chain, blockexplorer, indexers to own page

PR #402, opened 2024-09-18
docs: Add cheatsheet and improve organization

aefhm:
## Description

Add cheatsheet for hackathon. Follows https://github.com/oasisprotocol/sapphire-paratime/pull/400 and refactors "Contract Verification" to standalone page and blocks https://github.com/oasisprotocol/docs/pull/940.


rube-de (approved):

aefhm:
> if we have the verification as own page, I think the section in the guide can be removed at the same time

You are right.

rube-de (approved):

aefhm:
Preview does not work in this case because we update route in primary `docs` repository https://github.com/oasisprotocol/docs/blame/f9cbdd7f9eb7add92f5b79ad1b405688f6b15513/docs/dapp/tools/README.md#L9

matevz:
Thanks guys. Since Sourcify is an external service actually and not necessarily tied to Sapphire (emerald, pontus-x), should we move the new Contract verification page under the [Tools & Services](https://docs.oasis.io/dapp/tools/) part instead?

Issue #350, opened 2024-08-19
docs: Check out EIP-1167 and see how it fits in deployment chapter


matevz:
We got the following question:
> So it seems the contract is not verified, but we use a proxy for it, and the original contract is verified. Does your explorer support verification of proxy? We use the EIP1167 proxy for it. If not, is there a generic interface we could use to interact with the contract?

While the issue is primarily explorer and ABI playground related, it got me thinking why didn't we mention [EIP-1167](https://eips.ethereum.org/EIPS/eip-1167) in the [Deployment](https://docs.oasis.io/dapp/sapphire/deployment) chapter. Explore the EIP, build a minimal example, check out if this pattern can be used on Sapphire and document it.

aefhm:
I tried OpenZeppelin, but that borked. I think I need a different one. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/cae60c595b37b1e7ed7dd50ad0257387ec07c0cf/contracts/proxy/Clones.sol#L6

rube-de:
For me the OZ Clone worked https://github.com/rube-de/sapphire-minimal-proxy

the explorer got updates for it
https://github.com/oasisprotocol/explorer/pull/1538
https://github.com/oasisprotocol/explorer/pull/1540

thou it is only recognized after one transction on the proxy was executed.
with tx: https://explorer.oasis.io/testnet/sapphire/address/0xbA7F3eE55b7939294caFA0F1e096E79722e1BD88
w/o tx: https://explorer.oasis.io/testnet/sapphire/address/0x916336E6d9aE3225dEE3115d7803621022476d0e

aefhm:
Thank @rube-de. Got mine working as well.

PR #410, opened 2024-09-23
docs: Add EIP-1167 minimal proxy section

aefhm:
## Description

Close https://github.com/oasisprotocol/sapphire-paratime/issues/350 by noting full support.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -131,7 +131,25 @@ Singleton factory:
 [Testnet]: https://explorer.oasis.io/testnet/sapphire/address/0x914d7Fec6aaC8cd542e72Bca78B30650d45643d7
 [safe-singleton-factory]: https://github.com/safe-global/safe-singleton-factory
 
-### Caution Against Using `eth_getStorageAt`
+## Clones
+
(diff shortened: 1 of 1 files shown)


rube-de:
do we want to mention that the explorer will detect them?
or wait with it until it will also show verified?

aefhm:
> do we want to mention that the explorer will detect them?

I think this is expected behavior? Or do you mean that we should note that Explorers only detect after a proxy transaction?

> or wait with it until it will also show verified?

My first thought was to call out Sapphire differences only. 🤔

rube-de (approved):
Ok, I thinks it's fine like this. 
Initially I thought it could raise question for devs when they don't see it as proxy directly, but as the ticket for the Explorer is still open, I hope this will be fixed.

aefhm:
> Initially I thought it could raise question for devs when they don't see it as proxy directly,

I do find this to be an issue. I believe the other proxy contracts also aren't picked up correctly by the Explorer until a transaction is made. I wonder if this page or perhaps a new one is a better place for quirks with our Explorer.

> but as the ticket for the Explorer is still open, I hope this will be fixed.

Agree that the Explorer UI is the _**best**_ place to convey this information.

PR #411, opened 2024-09-24
clients: Support Brave wallet default

### LINKED ISSUE #385: sapphire-viem-v2: ProxyObject throws error due to changed function signature
##### SUMMARY

Using Brave wallet on Brave browser with the Viem client results in JS/TypeScript error. This was first reported by 0x2fa during testing. https://github.com/0x2fa-org/0x2fa

##### ISSUE TYPE
- Bug Report

##### STEPS TO REPRODUCE

Visit 0x2fa.com on Brave with default wallet selected of Brave wallet.

##### ACTUAL RESULTS

The way that the `sapphire-client` produces a Proxy object via wrapping the original Provider modifies the type.
https://github.com/oasisprotocol/sapphire-paratime/blob/aa5ea27f524138fbddf69c8c6217be51cdf983eb/clients/js/src/provider.ts#L197

Apparently, running a local application without binding the context works.

```JavaScript
Uncaught (in promise) TypeError: 'get' on proxy: property 'removeListener' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected 'function(e,t){var n,r,i,o,s;if(u(t),void 0===(r=this._events))return this;if(void 0===(n=r[e]))return this;if(n...<omitted>...s}' but got 'function () { [native code] }')
```

##### EXPECTED RESULTS

Brave wallet should work as a default on Brave browser.

aefhm:
## Description

Close #385 maybe.

## Problem

Wagmi is concerned with bad implementations of `EIP-1193` and wants to check that the `removeListener` function [exists](https://github.com/wevm/wagmi/blob/86c42248c2f34260a52ee85183c607315ae63ce8/packages/core/src/connectors/injected.ts#L327-L335). However, the Brave wallet overrides `ethereum` [properties](https://github.com/brave/brave-core/blob/74bf470a0291ea3719f1a75af066ee10b7057dbd/components/brave_wallet/resources/ethereum_provider.js#L13-L27) and does not like the way we currently use `Proxy` objects, giving a typeerror upon a `get` call.

Our current implementation is a classic Proxy [pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#no_private_property_forwarding) in order to preserve `this`.

```JavaScript
const proxy = new Proxy(aSecret, {
  get(target, prop, receiver) {
    const value = target[prop];
    if (value instanceof Function) {
      return function (...args) {
        return value.apply(this === receiver ? target : this, args);
      };
    }
    return value;
  },
});
```

simplified to:

https://github.com/oasisprotocol/sapphire-paratime/blob/42c1a69ee0bfc989cff1363ac661aeeef0f873d2/clients/js/src/provider.ts#L211

We can repro error via:

```JavaScript
class Secret {
  #x = 1;
  x() {
    return this.#x;
  }
}

const aSecret = new Secret();
Object.defineProperties(aSecret, {
  removeListener: {
    value: function () {},
    writable: false,
  }
});
const proxy = new Proxy(aSecret, {
  get(target, prop, receiver) {
    const value = target[prop];
    if (value instanceof Function) {
      return function (...args) {
        return value.apply(this === receiver ? target : this, args);
      };
    }
    return value;
  },
});
console.log(proxy.removeListener);
```

One solution is to not forward context in order to maintain the function signature for Brave browsers as noted in this PR.

### DIFF
--- a/clients/js/src/provider.ts
+++ b/clients/js/src/provider.ts
@@ -208,7 +208,17 @@ export function makeTaggedProxyObject<T extends object>(
       if (prop === propname) return options;
       if (hooks && prop in hooks) return Reflect.get(hooks, prop);
       const value = Reflect.get(upstream, prop);
-      return typeof value === 'function' ? value.bind(upstream) : value;
+
+      // Brave wallet web3provider properties are read only and throw typeerror
+      // https://github.com/brave/brave-core/blob/74bf470a0291ea3719f1a75af066ee10b7057dbd/components/brave_wallet/resources/ethereum_provider.js#L13-L27
+      // https://github.com/wevm/wagmi/blob/86c42248c2f34260a52ee85183c607315ae63ce8/packages/core/src/connectors/injected.ts#L327-L335
+      const propWritable =
+        Object.getOwnPropertyDescriptor(upstream, prop)?.writable !== false;
+
+      if (typeof value === 'function' && propWritable) {
+        return value.bind(upstream);
+      }
+      return value;
     },
   });
 }


aefhm:
One other solution mentioned would involve giving up the Proxy which currently prevents developers from accidentally double wrapping the `provider`. We ostensibly need that differentiation if but only for debugging purposes. Otherwise, I think creating a singleton factory would not necessarily be a clean solution, requiring a dev to continually reference the wrapped `SapphireProvider`.

aefhm:
Checking if @lukaw3d knows of a better magical JavaScript solution. 🙏🧙

CedarMist:
It should be possible to remove the proxy contract and use some kind of WeakRef singleton for providers instead?

Or... just don't do either.

As long as there's a test for using double-wrapped providers to make sure they don't mess up encryption & decryption.

aefhm:
Brave makes the upstream a [Proxy](https://github.com/brave/brave-core/blob/483894ebbdc17dbc36ca3e7d7e37ec69028166c9/components/brave_wallet/renderer/js_ethereum_provider.cc#L53-L57). 😆

CedarMist:
Can we do something like instead?

```typescript
return (...args) => Reflect.apply(value, upstream, ...args)
```

aefhm:
> Can we do something like instead?
> 
> ```ts
> return (...args) => Reflect.apply(value, upstream, args)
> ```
> 
> Which will avoid the `.bind` call

```JavaScript
Uncaught (in promise) TypeError: 'get' on proxy: property 'removeListener' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected 'function(e,t){var n,r,i,o,s;if(u(t),void 0===(r=this._events))return this;if(void 0===(n=r[e]))return this;if(n...<omitted>...s}' but got '(...args) => Reflect.apply(value, upstream, args)')
...
```

😭

clients/js/src/provider.ts
```diff
@@ -208,7 +215,16 @@ export function makeTaggedProxyObject<T extends object>(
...
+      // Brave wallet web3provider properties are read only and throw typeerror
+      // https://github.com/brave/brave-core/blob/74bf470a0291ea3719f1a75af066ee10b7057dbd/components/brave_wallet/resources/ethereum_provider.js#L13-L27
+      // https://github.com/wevm/wagmi/blob/86c42248c2f34260a52ee85183c607315ae63ce8/packages/core/src/connectors/injected.ts#L327-L335
+      if (
+        typeof value === 'function' &&
+        !BRAVE_READ_ONLY_PROPERTIES.includes(prop as string)
+      ) {
+        return value.bind(upstream);
+      }
+      return value;
```
lukaw3d, line 227: `if (Object.getOwnPropertyDescriptor(aSecret, 'removeListener').writable)`

aefhm:
> > Can we do something like instead?
> > ```ts
> > return (...args) => Reflect.apply(value, upstream, args)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which will avoid the `.bind` call
> 
> ```js
> Uncaught (in promise) TypeError: 'get' on proxy: property 'removeListener' is a read-only and non-configurable data property on the proxy target but the proxy did not return its actual value (expected 'function(e,t){var n,r,i,o,s;if(u(t),void 0===(r=this._events))return this;if(void 0===(n=r[e]))return this;if(n...<omitted>...s}' but got '(...args) => Reflect.apply(value, upstream, args)')
> ...
> ```
> 
> 😭

Still generates typeerror.

clients/js/src/provider.ts
aefhm: Yeah, this is better.

aefhm:
I think maintaining the Proxy object is a reasonable for now in order to support Brave or any other upstream that hijacks the `provider` with their `trap`. What do you think @CedarMist?

CedarMist (approved):

PR #419, opened 2024-09-27
integrations/viem: doctests

CedarMist:
Adds tests to ensure the README.md examples from integrations/viem-v2 can be imported.

This will catch any glaring errors. I think we should extend this to all of the packages in this repo.

### DIFF
--- a/integrations/viem-v2/README.md
+++ b/integrations/viem-v2/README.md
@@ -21,6 +21,7 @@ any unsigned communication, for example when using [hardhat-viem] pass the
 [hardhat-viem]: https://hardhat.org/hardhat-runner/docs/advanced/using-viem
 
 ```typescript
+import hre from "hardhat";
 import { sapphireLocalnet, sapphireHttpTransport } from '@oasisprotocol/sapphire-viem-v2';
 
 const publicClient = await hre.viem.getPublicClient({
@@ -35,7 +36,11 @@ transactions when using a local wallet client you must not only provide the
 `transport` parameter, but must also wrap the wallet client, as such:
 
 ```typescript
+import { createWalletClient } from 'viem'
+import { english, generateMnemonic, mnemonicToAccount } from 'viem/accounts';
 import { sapphireLocalnet, sapphireHttpTransport, wrapWalletClient } from '@oasisprotocol/sapphire-viem-v2';
+
+const account = mnemonicToAccount(generateMnemonic(english));
 
 const walletClient = await wrapWalletClient(createWalletClient({
 	account,
--- a/integrations/viem-v2/README.md
+++ b/integrations/viem-v2/README.md
@@ -2,38 +2,52 @@
 
 A plugin for [Viem] 2.x that encrypts transactions, gas estimations and calls to
 the Oasis Sapphire network to enable end-to-end encryption between the dApp and
-smart contracts.
+smart contracts. It provides three functions which can be used together or
+separately, see the guides below for usage instructions:
+
+ * `sapphireHttpTransport` - [Transport] that intercepts & encrypts requests.
+ * `wrapWalletClient` - Wraps a [WalletClient], to provide encryption.
+ * `createSapphireSerializer` - [Transaction serializer], that encrypts calldata.
 
 [Viem]: https://viem.sh/
+[WalletClient]: https://viem.sh/docs/clients/wallet.html
+[Transaction serializer]: https://viem.sh/docs/chains/serializers#api
+[Transport]: https://viem.sh/docs/clients/transports/http
 
 ## Usage
 
-First install the package.
+Add the package to your project:
 
 ```
 npm install @oasisprotocol/sapphire-viem-v2 viem@2.x
 ```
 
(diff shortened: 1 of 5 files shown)


integrations/viem-v2/README.md
rube-de, line 1: the readme uses hardhat-viem, `viem-v2.test.ts` doesn't.
I think we should have it consistent, either w/ hardhat-viem, w/o hardhat-viem or show both ways eveywhere.

integrations/viem-v2/package.json
```diff
@@ -59,7 +59,11 @@
 	},
 	"devDependencies": {
 		"@biomejs/biome": "^1.7.0",
+		"@nomicfoundation/hardhat-viem": "~2.0.0",
```
rube-de, line 62: whats the reason to fix it to patch version?
(beside that they never released a minor version)

integrations/viem-v2/README.md
CedarMist: The README has both with & without hardhat-viem.

integrations/viem-v2/package.json
CedarMist: We're not using hardhat-viem in this package at all, it's just to get the code example in README.md to work.

But yea, I should probably set it to ^2.0.0, but it's the default that came with `pnpm add -D '@nomicfoundation/hardhat-viem'`

integrations/viem-v2/README.md
rube-de: The README has
```js
const publicClient = await hre.viem.getPublicClient({
	chain: sapphireLocalnet,
	transport: sapphireHttpTransport()
});
```

the test has 
```js
const transport = sapphireHttpTransport();
const chain = sapphireLocalnet;
const publicClient = createPublicClient({ chain, transport });
```

also the test has for deployment
```js
const deployTxHash = await walletClient.deployContract({
	account,
	bytecode: ExampleContract.bytecode,
	abi: ExampleContract.abi,
});
const deployTxReceipt = await publicClient.waitForTransactionReceipt({
	hash: deployTxHash,
});
if (!deployTxReceipt.contractAddress) {
	console.log(deployTxReceipt);
	throw new Error("Contract not deployed");
}
const example = getContract({
	address: deployTxReceipt.contractAddress,
	abi: ExampleContract.abi,
	client: { public: publicClient, wallet: walletClient },
});
```

while you could write this with `hre.viem` to
```js
  const walletClient = await wrapWalletClient(createWalletClient({
      account, chain, transport
  }));
  keyedClient = {
      public: publicClient,
      wallet: walletClient
  }
  example = await hre.viem.deployContract('Example', [], {client:keyedClient});
```

CedarMist: I'm moving hardhat-viem stuff into a separate ticket so I can fully test it separately.
https://github.com/oasisprotocol/sapphire-paratime/issues/420

integrations/viem-v2/hardhat.config.cjs
rube-de, line 1: This file isn't need anymore, since hardhat isn't used

integrations/viem-v2/package.json
```diff
@@ -60,6 +60,9 @@
 	"devDependencies": {
 		"@biomejs/biome": "^1.7.0",
 		"@types/node": "^18.19.31",
+		"@zaeny/literate": "^1.0.8",
+		"hardhat": "^2.22.12",
```
rube-de, line 64:
```suggestion
```
Since hardhat isn't used anymore this dependency can be removed

lukaw3d:
Nice!

---

If we'll need snippets that skip some initialization code - copying a comment from slack:
For previous attempts at testing code in markdown we pulled code in from tests:
- generated documentation using vuepress snippets https://v1.vuepress.vuejs.org/guide/markdown.html#import-code-snippets
  (https://github.com/vuejs/vuepress/issues/2663)
- docs.oasis.io generated using docusaurus + https://github.com/oasisprotocol/docs/blob/ded70ed/src/remark/code-block-snippets.ts

  https://docs.oasis.io/rofl/app#testing-it-on-sapphire-localnet 
  -> https://github.com/oasisprotocol/oasis-sdk/blob/d23473d/docs/rofl/app.md?plain=1#L122
  -> https://github.com/oasisprotocol/oasis-sdk/blob/d23473d/examples/runtime-sdk/rofl-oracle/src/main.rs#L24-L35

New idea that would work with npm and github markdown without generating them:
- keep code as is in markdown
- add same code to files in tests
- add test that checks that code is found in a test (ignore indentation/whitespace)

integrations/viem-v2/package.json
```diff
@@ -60,6 +60,8 @@
 	"devDependencies": {
 		"@biomejs/biome": "^1.7.0",
 		"@types/node": "^18.19.31",
+		"@zaeny/literate": "^1.0.8",
```
lukaw3d, line 63: Maybe a license issue

CedarMist: You're right. It's silly to use an external package when it's basically just a single regex. Have rewritten to avoid the external package.

rube-de (approved):

Issue #421, opened 2024-09-29
Go client doesn't use core.CallDataPublicKey


CedarMist:
This means if Docker runs on a different port or IP address the retrieval will fail

PR #422, opened 2024-09-29
clients/go: use existing RPC connection to retrieve calldata public key

CedarMist:
fixes: #421
adds tests for performing a value transfer (doesn't check if encryption works)

### DIFF
--- a/.github/workflows/ci-test-go.yaml
+++ b/.github/workflows/ci-test-go.yaml
@@ -1,0 +1,43 @@
+name: ci-test
+
+on:
+  push:
+    branches:
+      - main
+      - stable/*
+      - rc/*
+  pull_request:
+    branches:
+      - main
+      - stable/*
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-test-go.yaml
```diff
@@ -0,0 +1,43 @@
+name: ci-test
+
+on:
+  push:
+    branches:
+      - main
+      - stable/*
+      - rc/*
```
aefhm, line 8: Do we need this distinction? I understand this is a non-unique change to this PR.

CedarMist: I just copied the action and added the sapphire localnet container, given we will likely be changing the actions in another PR to make CI run quicker and the status badges etc more accurately represent the different components I say we address that later

abukosek (approved):

PR #431, opened 2024-10-09
Initial work to implement Sapphire snap connection

### LINKED ISSUE #389: sapphire-snap integration
We need to detect if the `sapphire-snap` is installed.

The RPC call `wallet_getSnaps'`, and check if sapphire-snap is loaded.

If it's loaded, then we can do the `wallet_invokeSnap` for the `setTransactionDecryptKeys` method.

CedarMist:
fixes #389

This provides the decryption keys to snap.

> [!WARNING]
> If an RPC server pretends to implement the MetaMask snap protocol it could trick users into revealing the transaction encryption key. 
>
> For this reason, we have to explicitly enable Snap support in the dApp, by passing the `enableSapphireSnap` option.

Usage:

```typescript
wrapEthereumProvider(window.ethereum, {enableSapphireSnap:true})
```

```typescript
wrapEthersProvider(ethProvider, { enableSapphireSnap: true })
```

This must only be done if the dApp is sure that the provider it's connecting to is MetaMask.

Next PRs handled by Sapphire team: 

 * Automated tests for `wallet_invokeSnap`
 * Use flask for testing https://docs.metamask.io/snaps/get-started/install-flask/

### DIFF
--- a/clients/js/src/cipher.ts
+++ b/clients/js/src/cipher.ts
@@ -216,7 +216,7 @@ export class X25519DeoxysII extends Cipher {
   public override readonly epoch: number | undefined;
 
   private cipher: deoxysii.AEAD;
-  private key: Uint8Array; // Stored for curious users.
+  public secretKey: Uint8Array; // Stored for curious users.
 
   /** Creates a new cipher using an ephemeral keypair stored in memory. */
   static ephemeral(peerPublicKey: BytesLike, epoch?: number): X25519DeoxysII {
@@ -254,8 +254,8 @@ export class X25519DeoxysII extends Cipher {
       .update(naclScalarMult(keypair.secretKey, peerPublicKey))
       .digest().buffer;
 
-    this.key = new Uint8Array(keyBytes);
-    this.cipher = new deoxysii.AEAD(new Uint8Array(this.key)); // deoxysii owns the input
+    this.secretKey = new Uint8Array(keyBytes);
+    this.cipher = new deoxysii.AEAD(new Uint8Array(this.secretKey)); // deoxysii owns the input
   }
 
   public encrypt(
--- a/clients/js/src/cipher.ts
+++ b/clients/js/src/cipher.ts
@@ -80,6 +80,7 @@ function formatFailure(fail: CallFailure): string {
 export abstract class Cipher {
   public abstract kind: CipherKind;
   public abstract publicKey: Uint8Array;
+  public abstract ephemeralKey: Uint8Array;
   public abstract epoch?: number;
 
   public abstract encrypt(
@@ -213,6 +214,7 @@ export abstract class Cipher {
 export class X25519DeoxysII extends Cipher {
   public override readonly kind = CipherKind.X25519DeoxysII;
   public override readonly publicKey: Uint8Array;
+  public override readonly ephemeralKey: Uint8Array;
   public override readonly epoch: number | undefined;
 
   private cipher: deoxysii.AEAD;
@@ -243,6 +245,7 @@ export class X25519DeoxysII extends Cipher {
     super();
 
     this.publicKey = keypair.publicKey;
+    this.ephemeralKey = keypair.secretKey;
     this.epoch = epoch;
 
     // Derive a shared secret using X25519 (followed by hashing to remove ECDH bias).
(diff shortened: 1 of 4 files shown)


clients/js/src/provider.ts
```diff
@@ -43,6 +44,7 @@ export function isLegacyProvider<T extends object>(
 
 export interface SapphireWrapOptions {
   fetcher: KeyFetcher;
```
buberdds, line 46: Options object is optional. When provided, TS requires fetcher, but we should be able just to set `enableSapphireSnap` flag. As fetcher always defaults to `new KeyFetcher() ` should it now be `fetcher: KeyFetcher | undefined;` ?

clients/js/src/provider.ts
```diff
@@ -80,10 +82,10 @@ export function isWrappedEthereumProvider<P extends EIP2696_EthereumProvider>(
  * @param options (optional) Re-use parameters from other providers
  * @returns Sapphire wrapped provider
  */
-export function wrapEthereumProvider<P extends EIP2696_EthereumProvider>(
+export async function wrapEthereumProvider<P extends EIP2696_EthereumProvider>(
```
buberdds, line 85: Does it affect all integrations? For example I don't see ether-v6 is relying on `wrapEthereumProvider`. It has it's own logic.

clients/js/src/provider.ts
```diff
@@ -124,6 +126,63 @@ export function wrapEthereumProvider<P extends EIP2696_EthereumProvider>(
...
+    await provider.request({
+      method: 'wallet_invokeSnap',
+      params: {
+        snapId: snapId,
+        request: {
+          method: 'setTransactionDecryptKeys',
+          params: {
+            id: transactionData,
+            ephemeralSecretKey: hexlify(secretKey),
+            peerPublicKey: peerPublicKey.key,
```
buberdds, line 177: hexlify

clients/js/src/cipher.ts
```diff
@@ -213,10 +214,11 @@ export abstract class Cipher {
 export class X25519DeoxysII extends Cipher {
   public override readonly kind = CipherKind.X25519DeoxysII;
   public override readonly publicKey: Uint8Array;
+  public override readonly ephemeralKey: Uint8Array;
   public override readonly epoch: number | undefined;
 
   private cipher: deoxysii.AEAD;
-  private key: Uint8Array; // Stored for curious users.
+  public secretKey: Uint8Array; // Stored for curious users.
```
buberdds, line 221: That's not a key we need, added `ephemeralkey`.

clients/js/src/provider.ts
```diff
@@ -99,7 +101,7 @@ export function wrapEthereumProvider<P extends EIP2696_EthereumProvider>(
   // if we do this, don't then re-wrap the send() function
   // only wrap the send() function if there was a request() function
 
-  const request = makeSapphireRequestFn(upstream, filled_options);
+  const request = await makeSapphireRequestFn(upstream, filled_options);
```
buberdds: I have reverted a change of making it all async, because async part can be handled in proxy

matevz (commented):
First pass. I wonder whether we should be concerned about other JS scripts (e.g. maliciously injected on a website) accessing the `ephemeralKey` field and extracting secret material. Is `private` vs `public` visibility of javascript vars playing an important part here?

clients/js/src/cipher.ts
buberdds: Derived shared secret key from ephemeral and peer public key. I am not sure what are all use cases looking at comment, but it's not a key needed to init cipher via `X25519DeoxysII.fromSecretKey`

clients/js/src/cipher.ts
```diff
@@ -79,7 +79,11 @@ function formatFailure(fail: CallFailure): string {
 
 export abstract class Cipher {
   public abstract kind: CipherKind;
+  // The public part of the ephemeral key pair, rotated for each epoch.
   public abstract publicKey: Uint8Array;
+  // The private part of the ephemeral key pair, rotated for each epoch.
+  // Used in combination with a peer's public key for encryption/decryption using the X25519-DeoxysII.
```
matevz, line 85: This reads wrong. Client doesn't have access to Sapphire's private ephemeral key. The client only has access to the Sapphire's public key and then the client's own public/private keypair. The latter is ephemeral as well, but is rotated on-demand, no need to be limited by the epoch.

I would say sth like `publicKey` is the Sapphire's public key rotated each epoch. And `ephemeralKey` is the client's keypair for encrypting/decrypting transactions and queries using the X25519-DeoxysII, rotated on-demand.

matevz (approved):
Please check my comment again. Otherwise looks good.

Issue #428, opened 2024-10-08
Support generating encrypted transactions on-chain


matevz:
The current EIP155Signer expects arbitrary calldata, e.g. https://docs.oasis.io/dapp/sapphire/gasless#gasless-proxy-contract.

Add a helper to Sapphire.sol or other contract for on-chain [Deoxys-II encryption](https://api.docs.oasis.io/sol/sapphire-contracts/contracts/Sapphire.sol/library.Sapphire.html#encrypt-1) of the call data prior to signing the transaction. See how to fetch the ephemeral public key [here](https://github.com/oasisprotocol/sapphire-paratime/pull/370).

Update the [Gasless chapter](https://docs.oasis.io/dapp/sapphire/gasless#on-chain-signer) with the new encrypted version of the tx.

CedarMist:
It may be worth extracting the CBOR related functions into a helper library, although I'd be wary of modifying or refactoring of the existing functions.

CedarMist:
Made a quick sketch so far, will test it

```solidity
// SPDX-License-Identifier: Apache-2.0

pragma solidity ^0.8.0;

import { HMAC_sha512_256 } from './HMAC_sha512_256.sol';
import { Subcall } from '@oasisprotocol/sapphire-contracts/contracts/Subcall.sol';
import { Sapphire } from '@oasisprotocol/sapphire-contracts/contracts/Sapphire.sol';

library EncryptedTx {
    function internal_deriveSharedSecret(
        bytes32 peerPublicKey,
        Sapphire.Curve25519SecretKey mySecretKey
    )
        internal view
        returns (bytes32)
    {
        return HMAC_sha512_256(
            "MRAE_Box_Deoxys-II-256-128",
            abi.encodePacked(
                Sapphire.deriveSymmetricKey(
                    Sapphire.Curve25519PublicKey.wrap(peerPublicKey),
                    mySecretKey)));
    }

    error CBORBytesTooLong();

    function _cborEncodeBytes(bytes memory in_bytes)
        internal pure
        returns (bytes memory out_cbor)
    {
        /*
        0x40..0x57 	byte string (0x00..0x17 bytes follow)
        0x58 	byte string (one-byte uint8_t for n, and then n bytes follow)
        0x59 	byte string (two-byte uint16_t for n, and then n bytes follow)
        0x5a 	byte string (four-byte uint32_t for n, and then n bytes follow)
[...]

PR #436, opened 2024-10-11
contracts: implement calldata encryption in Solidity

CedarMist:
Fixes: #428

 * HMAC SHA512-256 (but discovered that was unnecessary. Included anyway as it may be useful.)
 * Introduces `subcall_static` to perform subcall with `staticcall`, makes several Subcall methods `view`
 * Split CBOR stuff into its own file + tests.
 * Allow passing manual `nonce` to Cipher for encryption (useful for tests & reproducibility)
 * Various changes to clean up solhint warnings
 * Transaction calldata encryption + tests

Re: CBOR, I split the CBOR related stuff into individual functions, as otherwise you start needing to deploy the whole `CBOR` library on its own an link it into a library, whereas using individual function they get pulled in as & when necessary.


contracts/contracts/CalldataEncryption.sol
```diff
@@ -0,0 +1,118 @@
...
+
+        bytes memory inner = _encryptInner(
+            in_data,
+            mySecret,
+            nonce,
+            peerPublicKey
+        );
+
+        return
+            abi.encodePacked(
```
CedarMist, line 94: I suspect the output bytecode for this is going to be huge

contracts/test/cbor.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    );
+
+    const bytes = cborg.encode(uint128bytes);
+
+    const [newOffset, parsedCborUint] = await contract.testParseUint(bytes, 0);
+
+    expect(parsedCborUint).eq(MAX_SAFE_UINT128);
+    expect(newOffset).eq(16 + 1);
+
+    //expect(await contract.testUintRoundtrip(MAX_SAFE_UINT128)).eq(true);
```
CedarMist, line 108: CBOR encoding integers larger than 64bit seems to vary between libraries, some use tagging, others don't.

contracts/test/calldata.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+        peerKeypair.publicKey,
+      );
+      expect(result).eq(hexlify(encryptedCall));
+    }
+  });
+
+  it('roundtrip encryption', async () => {
+    const myAddr = '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266';
+    const myKey =
+      '0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80';
```
CedarMist, line 65: TODO: verify `ethers.getSigner(0)` equals `myAddr`

contracts/test/calldata.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    for (let i = 1; i < 1024; i += 1 + i / 5) {
+      const peerKeypair = boxKeyPairFromSecretKey(
+        randomBytes(crypto_box_SECRETKEYBYTES),
+      );
+      const myKeypair = boxKeyPairFromSecretKey(
+        randomBytes(crypto_box_SECRETKEYBYTES),
+      );
+      const calldata = randomBytes(i);
+      const epoch = randomInt(1 << 32);
+      const nonce = randomBytes(15);
```
CedarMist, line 43: Surely nonce length should be pulled from somewhere else? Instead of defined as `15`

contracts/contracts/CalldataEncryption.sol
```diff
@@ -0,0 +1,118 @@
...
+
+    function _encryptInner(
+        bytes memory in_data,
+        Sapphire.Curve25519SecretKey in_x25519_secret,
+        bytes15 nonce,
+        bytes32 peerPublicKey
+    ) internal view returns (bytes memory out_encrypted) {
+        bytes memory plaintextEnvelope = abi.encodePacked(
+            hex"a1", // map(1)
+            hex"64", //     text(4) "body"
```
aefhm, line 29:
```suggestion
            hex"64", // text(4) "body"
```

contracts/contracts/tests/TestCBOR.sol
```diff
@@ -0,0 +1,40 @@
...
+        external pure
+        returns (bool)
+    {
+        bytes memory encoded = CBOR_encodeUint(value);
+        (uint newOffset, uint result) = CBOR_parseUint(encoded, 0);
+        require( result == value, "value wrong!" );
+        require( newOffset == encoded.length, "didn't parse everything!" );
+        return true;
+    }
+}
```
aefhm, line 40:
```suggestion
}

```

contracts/test/cbor.spec.ts
aefhm: Not a blocker right. I'm sure we'll get dev feedback if this is buggy?

contracts/test/calldata.spec.ts
aefhm: Yeah, this could also be slightly surprising?

contracts/contracts/CBOR.sol
rube-de, line 1: I would import the free functions and error with `import ./CBOR.sol as CBOR` and drop the CBOR prefixing of the functions. 
The CBOR prefix for errors makes sense as this indicates where the error comes from, but I think the error in the naming is self-explanatory and could be dropped

contracts/contracts/CalldataEncryption.sol
```diff
@@ -0,0 +1,118 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pragma solidity ^0.8.0;
+
+import {Subcall} from "./Subcall.sol";
+import {Sapphire} from "./Sapphire.sol";
+import {CBOR_encodeBytes, CBOR_encodeUint} from "./CBOR.sol";
```
rube-de, line 7:
```suggestion
import "./CBOR.sol" as CBOR;
```
according to the [announcement of free functions](https://soliditylang.org/blog/2020/09/02/solidity-0.7.1-release-announcement/), "it is recommended to use the import statement that generates a new object, so that the file's scope is not polluted needlessly"

This would also come with prefixing all calls to the free functions with CBOR.
Which then probably makes prefixing the functions in CBOR.sol obsolete.

contracts/contracts/CBOR.sol
CedarMist: That's a really good point, I'll keep the CBOR prefix for errors, but remove it from the methods themselves.

contracts/contracts/CalldataEncryption.sol
CedarMist: This is formatted according to the CBOR debug output I'm using throughout this PR.
It's vaguely hierarchial.
Unfortunately the Solidity prettifier is butchering any attempt other than comments to make it readable.

clients/js/src/cipher.ts
```diff
@@ -104,7 +110,8 @@ export abstract class Cipher {
 
     const innerEnvelope = cborEncode({ body: getBytes(calldata) });
 
-    const { ciphertext, nonce } = this.encrypt(innerEnvelope);
+    let ciphertext: Uint8Array;
+    ({ ciphertext, nonce } = this.encrypt(innerEnvelope, nonce));
```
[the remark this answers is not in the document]
CedarMist: The nonce can be specified by the user.
If it's not, then `encrypt` will generate one for them.
that's then used below in the envelope.

contracts/test/calldata.spec.ts
CedarMist: Nonce length exported by the deoxysii package, but not by the Cipher object.

I think just leave this for now as it's in tests and if anything changes it'll break and be easy to find.

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,169 @@
...
+    }
+    // Unknown...
+    else {
+        revert CBOR_Error_InvalidUintPrefix(prefix);
+    }
+
+    if (len > 0x20) revert CBOR_Error_InvalidLength(len);
+
+    assembly {
+        value := mload(add(add(0x20, result), offset))
```
matevz, line 122: Huh?

contracts/contracts/OPL.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
-import "./opl/Endpoint.sol";
-import "./opl/Enclave.sol";
-import "./opl/Host.sol";
+import {ICelerMessageBus, BaseEndpoint, Endpoint, autoswitch} from "./opl/Endpoint.sol";
```
matevz, line 4: Is this related to calldata encryption?

matevz (commented):
I know I'm PITA, but this is tedious to review. Why not make life easier and split this PR into five:
- CBOR.sol
- actual calldata encryption PR
- OPL support for encrypted calldata
- HMAC_sha512_256
- 0.3.0 version bump

contracts/contracts/CBOR.sol
CedarMist: This loads a 256bit word from an offset of a byte array, the first 256bits of the byte array is the length prefix, which is skipped.

CedarMist:
Unfortunately splitting a small PR into lots of even smaller minor PRs is also very tedious. I can re-commit them as individual commits within the PR (that would be slightly less tedious)?

contracts/contracts/CBOR.sol
matevz: Ah, I see. Maybe add a comment?

```suggestion
        // Load value from the result array at position 32+offset.
        value := mload(add(add(0x20, result), offset))
```

Issue #430, opened 2024-10-08
Describe length of Sapphire.generateSigningKeyPair private key


matevz:
A user [asked](https://discord.com/channels/748635004384313474/960599114993762304/1293144981124681728) about the length of the private key of generateSigningKeyPair.

The https://api.docs.oasis.io/sol/sapphire-contracts/contracts/Sapphire.sol/library.Sapphire.html#generatesigningkeypair page is missing the length of the returned private key. Add it.

PR #437, opened 2024-10-14
contracts/generateSigningKeypair: Doc explicit sk size

matevz:
Fixes https://github.com/oasisprotocol/sapphire-paratime/issues/430

### DIFF
--- a/contracts/contracts/Sapphire.sol
+++ b/contracts/contracts/Sapphire.sol
@@ -321,16 +321,27 @@ library Sapphire {
      * ##### Secp256r1: 4,000 gas
      * - `7` (`Secp256r1PrehashedSha256`)
      *
-     * #### Public Key Format
+     * ##### Secp384r1: 18,000 gas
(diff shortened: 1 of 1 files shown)


CedarMist (changes requested):
With P384 the public/secret key sizes will be larger than 33 bytes, but follows the same encoding as other non-Edwards curves.

P384 = 49 bytes encoded (48 + 1 byte prefix)

Gas cost for P384:

 * Keygen: 18000
 * Sign: 43200
 * Verify: 37920

matevz:
> With P384 the public/secret key sizes will be larger than 33 bytes, but follows the same encoding as other non-Edwards curves.
> 
> P384 = 49 bytes encoded (48 + 1 byte prefix)
> 
> Gas cost for P384:
> 
>     * Keygen: 18000
> 
>     * Sign: 43200
> 
>     * Verify: 37920

Thanks, added P384 gas costs.

CedarMist (approved):

PR #438, opened 2024-10-15
contracts: minor fixes to reduce warnings from solhint

CedarMist:


contracts/.prettierignore
```diff
@@ -1 +1,2 @@
 typechain-types/
+contracts/tests/
```
matevz, line 2: Leftover?

matevz (approved):

contracts/.prettierignore
CedarMist: Prettier fights with solhint, I couldn't find a workaround other than asking prettier to ignore it. Otherwise it gets stuck in infinite loop of prettier moving the solhint comments to the wrong line, which re-enables the previous lint warning and potentially dangerously ignores an unrelated line of code. (Although in this case they're just empty function bodies in the tests)

PR #439, opened 2024-10-15
contracts: implement HMAC SHA512-256, as it's used throughout the Oasis ecosystem

CedarMist:


contracts/contracts/HMAC_sha512_256.sol
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * #### Example
+ *
+ * ```solidity
+ * bytes memory key = "arbitrary length key";
+ * bytes memory message = "arbitrary length message";
+ * bytes32 hmac = HMAC_sha512_256(key, message)
+ * ```
+ */
+function HMAC_sha512_256(bytes memory key, bytes memory message)
```
matevz, line 33: Was thinking about the function name. Solidity suggests mixed case for functions: https://docs.soliditylang.org/en/latest/style-guide.html#function-names

But since we're already having the underscore `sha512_256` then name it `hmac_sha512_256`? I would avoid upper cases.

```suggestion
function hmac_sha512_256(bytes memory key, bytes memory message)
```

CedarMist: I think HMAC_SHA512_256 is a better name, it is more common to use upper case letters to form acronyms.

contracts/package.json
```diff
@@ -53,7 +53,8 @@
     "solidity-coverage": "^0.8.2",
     "ts-node": "^10.9.1",
     "typechain": "^8.3.2",
-    "typescript": "^4.8.3"
+    "typescript": "^4.8.3",
+    "@noble/hashes": "1.3.2"
```
CedarMist, line 57: Note that v1.3.2 is used explicitly here, as that's what ethers depends on.

I guess we could use ^1.3.2 as latest viem uses 1.5.0

contracts/contracts/HMAC_sha512_256.sol
```diff
@@ -0,0 +1,70 @@
...
+    returns (bytes32)
+{
+    bytes32[4] memory buf;
+
+    if (key.length > SHA512_256_BLOCK_SIZE) {
+        buf[0] = sha512_256(key);
+    } else {
+        bool success;
+
+        assembly {
```
matevz, line 44: Please document this. It's about copying `size` bytes from `key` to `buf`? Is this memory safe?

matevz (approved):

contracts/contracts/HMAC_sha512_256.sol
CedarMist: That is a very good question.

Lets create an issue for it, then at the next sprint meeting we can decide if we go ahead with the rename. Then I can submit a PR for it.

CedarMist: Some notes for future discussin: renaming a variable of a function which has already been used in code will cause annoying breakages. We should defer the rename to sapphire-contracts 1.x so as to maintain source compatibility

contracts/contracts/HMAC_sha512_256.sol
CedarMist: Hmm, I'm not sure the updated comments cover everything, there are many aspects which have been omitted for brevity.

CedarMist:
I forgot to run `format` on the code after adding comments.

For clarity, which style do you prefer:

```diff
diff --git a/contracts/contracts/HMAC_sha512_256.sol b/contracts/contracts/HMAC_sha512_256.sol
index 119f907..06becfc 100644
--- a/contracts/contracts/HMAC_sha512_256.sol
+++ b/contracts/contracts/HMAC_sha512_256.sol
@@ -51,12 +51,12 @@ function HMAC_sha512_256(bytes memory key, bytes memory message)
             let size := mload(key)
             // Call the identity precompile to copy memory
             success := staticcall(
-                gas(),           // Forward all available gas
-                PRECOMPILE_IDENTITY_ADDRESS,  // Address of the identity precompile
-                add(32, key),    // Start of the key data (skip the length prefix)
-                size,            // Length of data to copy
-                buf,             // Destination to copy to
-                size             // Amount of memory to copy
+                gas(), // Forward all available gas
+                PRECOMPILE_IDENTITY_ADDRESS, // Address of the identity precompile
+                add(32, key), // Start of the key data (skip the length prefix)
+                size, // Length of data to copy
+                buf, // Destination to copy to
+                size // Amount of memory to copy
             )
         }
 
```

CedarMist:
The auth tests are still causing intermittent failures

contracts/contracts/HMAC_sha512_256.sol
```diff
@@ -0,0 +1,78 @@
...
+            )
+        }
+
+        // Ensure the memory copy was successful
+        require(success, "memcpy");
+    }
+
+    for (uint256 i = 0; i < buf.length; i++) {
+        buf[i] ^= HMAC_IPAD;
+    }
```
rube-de, line 69:
```suggestion
    buf[0] ^= HMAC_IPAD;
    buf[1] ^= HMAC_IPAD;
    buf[2] ^= HMAC_IPAD;
    buf[3] ^= HMAC_IPAD;
```
as this is a small fixed size array, unrolling the loop would save gas

contracts/contracts/HMAC_sha512_256.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    bytes32 ihash = sha512_256(abi.encodePacked(buf, message));
+
+    for (uint256 i = 0; i < buf.length; i++) {
+        buf[i] ^= HMAC_OPAD_XOR_IPAD;
+    }
```
rube-de, line 75:
```suggestion
    buf[0] ^= HMAC_OPAD_XOR_IPAD;
    buf[1] ^= HMAC_OPAD_XOR_IPAD;
    buf[2] ^= HMAC_OPAD_XOR_IPAD;
    buf[3] ^= HMAC_OPAD_XOR_IPAD;
```
as this is a small fixed size array, unrolling the loop would save gas

contracts/test/hashes.ts
```diff
@@ -41,4 +37,18 @@ describe('Hashes', () => {
   it('SHA384', async () => {
     await testHashes('SHA384', contract.testSHA384.bind(contract));
   });
+
+  it('HMAC SHA512-256', async () => {
+    for (let i = 0; i < 1024; i = i + (1 + i / 5)) {
```
matevz, line 42: Is there a specific reason to use `/ 5`?

matevz (approved):
Comments are too verbose IMO. Otherwise looks good.

contracts/test/hashes.ts
CedarMist: Because it's a nice sequence that tests boundaries key length and varying message sizes, without testing too many of them. I'm unaware of any standard test vectors for HMAC SHA512-256.

CedarMist:
Auth tests are still intermittently failing

matevz:
> SIWE auth tests are still intermittently failing

I think I found the issue https://github.com/oasisprotocol/sapphire-paratime/issues/443. Will fix it

matevz (approved):
Nice!

rube-de (approved):

PR #440, opened 2024-10-15
contracts: move CBOR related encoding & decoding into separate module

CedarMist:
as per @matevz 's request, I am splitting https://github.com/oasisprotocol/sapphire-paratime/pull/436 into several smaller ones for clarity.

### DIFF
--- a/contracts/contracts/CBOR.sol
+++ b/contracts/contracts/CBOR.sol
@@ -1,0 +1,172 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pragma solidity ^0.8.0;
+
+/// While parsing CBOR map, unexpected key
+error CBOR_Error_InvalidKey();
+
+/// While parsing CBOR map, length is invalid, or other parse error
+error CBOR_Error_InvalidMap();
+
+/// While parsing CBOR structure, data length was unexpected
+error CBOR_Error_InvalidLength(uint256);
+
+/// Value cannot be parsed as a uint
+error CBOR_Error_InvalidUintPrefix(uint8);
+
+/// Unsigned integer of unknown size
+error CBOR_Error_InvalidUintSize(uint8);
+
+/// CBOR parsed value is out of expected range
+error CBOR_Error_ValueOutOfRange();
+
+error CBOR_Error_BytesTooLong(uint256 byteLength);
+
+function encodeUint(uint256 value) pure returns (bytes memory) {
+    // NOTE: we don't follow bignum tagged encoding
+    // See: https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.3
+    if (value < 24) {
+        return abi.encodePacked(uint8(value));
+    } else if (value <= type(uint8).max) {
+        return abi.encodePacked(uint8(24), uint8(value));
+    } else if (value <= type(uint16).max) {
+        return abi.encodePacked(uint8(25), uint16(value));
+    } else if (value <= type(uint32).max) {
+        return abi.encodePacked(uint8(26), uint32(value));
+    } else if (value <= type(uint64).max) {
+        return abi.encodePacked(uint8(27), uint64(value));
+    } else if (value <= type(uint128).max) {
+        return abi.encodePacked(uint8(0x50), uint128(value));
+    }
+    return abi.encodePacked(uint8(0x58), uint256(32), value);
+}
+
+function encodeBytes(bytes memory in_bytes)
+    pure
+    returns (bytes memory out_cbor)
+{
+    /*
+    0x40..0x57 	byte string (0x00..0x17 bytes follow)
+    0x58 	byte string (one-byte uint8_t for n, and then n bytes follow)
+    0x59 	byte string (two-byte uint16_t for n, and then n bytes follow)
+    0x5a 	byte string (four-byte uint32_t for n, and then n bytes follow)
+    0x5b 	byte string (eight-byte uint64_t for n, and then n bytes follow)
+    */
+    if (in_bytes.length <= 0x17) {
+        return abi.encodePacked(uint8(0x40 + in_bytes.length), in_bytes);
+    }
+    if (in_bytes.length <= 0xFF) {
+        return abi.encodePacked(uint8(0x58), uint8(in_bytes.length), in_bytes);
+    }
+    if (in_bytes.length <= 0xFFFF) {
+        return abi.encodePacked(uint8(0x59), uint16(in_bytes.length), in_bytes);
+    }
+    // We assume Solidity won't be encoding anything larger than 64kb
+    revert CBOR_Error_BytesTooLong(in_bytes.length);
+}
+
+function parseMapStart(bytes memory in_data, uint256 in_offset)
+    pure
+    returns (uint256 n_entries, uint256 out_offset)
+{
+    uint256 b = uint256(uint8(in_data[in_offset]));
+    if (b < 0xa0 || b > 0xb7) {
+        revert CBOR_Error_InvalidMap();
+    }
+
+    n_entries = b - 0xa0;
+    out_offset = in_offset + 1;
+}
+
+function parseUint(bytes memory result, uint256 offset)
+    pure
+    returns (uint256 newOffset, uint256 value)
+{
+    uint8 prefix = uint8(result[offset]);
+    uint256 len;
+
+    if (prefix <= 0x17) {
+        return (offset + 1, prefix);
+    }
+    // Byte array(uint256), parsed as a big-endian integer.
+    else if (prefix == 0x58) {
+        len = uint8(result[++offset]);
+        offset++;
+    }
+    // Byte array, parsed as a big-endian integer.
+    else if (prefix & 0x40 == 0x40) {
+        len = uint8(result[offset++]) ^ 0x40;
+    }
+    // Unsigned integer, CBOR encoded.
+    else if (prefix & 0x10 == 0x10) {
+        if (prefix == 0x18) {
+            len = 1;
+        } else if (prefix == 0x19) {
+            len = 2;
+        } else if (prefix == 0x1a) {
+            len = 4;
+        } else if (prefix == 0x1b) {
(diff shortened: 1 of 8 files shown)


contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,171 @@
...
+/// While parsing CBOR structure, data length was unexpected
+error CBOR_Error_InvalidLength(uint256);
+
+/// Value cannot be parsed as a uint
+error CBOR_Error_InvalidUintPrefix(uint8);
+
+/// Unsigned integer of unknown size
+error CBOR_Error_InvalidUintSize(uint8);
+
+/// CBOR parsed valid is out of expected range
```
matevz, line 20:
```suggestion
/// CBOR parsed value is out of expected range
```

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,171 @@
...
+error CBOR_Error_ValueOutOfRange();
+
+error CBOR_Error_UintTooLong();
```
matevz, line 23: Missing doc.

```suggestion
/// The size of Uint exceeds 8 bytes
error CBOR_Error_UintTooLong();
```

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,171 @@
...
+
+error CBOR_Error_BytesTooLong();
```
matevz, line 25: Missing doc.

```suggestion
/// Too many bytes
error CBOR_Error_BytesTooLong();
```

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,171 @@
...
+    } else if (value <= type(uint8).max) {
+        return abi.encodePacked(uint8(24), uint8(value));
+    } else if (value <= type(uint16).max) {
+        return abi.encodePacked(uint8(25), uint16(value));
+    } else if (value <= type(uint32).max) {
+        return abi.encodePacked(uint8(26), uint32(value));
+    } else if (value <= type(uint64).max) {
+        return abi.encodePacked(uint8(27), uint64(value));
+    }
+    // XXX: encoding beyond 64bit uints isn't 100% supported
```
matevz, line 39: Smells more like TODO than dirty code ;)

```suggestion
    // TODO: encoding beyond 64bit uints isn't 100% supported
```

contracts/contracts/CBOR.sol
CedarMist: This is a "redundant comment" or more specifically, "comment-name redundancy."

The symbol should be renamed instead.

Which one should we choose, CBOR_Error_TooManyBytes or CBOR_Error_BytesTooLong, which name makes most sense within the context that it's thrown. Perhaps @matevz could even suggest a better error name.

contracts/contracts/CBOR.sol
CedarMist: This could also be considered a redundant comment, I haven't decided yet if it's worth implementing uint parsing for 128bit and 256bits.

Accepting this comment would make it more likely for the comments to become out of sync with functionality.

Lets decide, should we implement uint128 and uint256 parsing, or only handle 8,16,32 & 64 bits?

contracts/contracts/CBOR.sol
matevz: I would get rid of it and support 0x5a and 0x5b types too in `encodeBytes()` below.

contracts/contracts/CBOR.sol
matevz: Implementing uint128 and uint256 parsing would make sense, if there is a solidity<->solidity communication happening between the contracts in CBOR.

CedarMist: The `cborg` library we use in Javascript doesn't support encoding/decoding of bigints beyond 64bit.

https://github.com/rvagg/cborg/blob/eca76d557991a4ded99bfe8ed062089b7bc5b21e/lib/0uint.js#L187

There is no Solidity <-> Solidity communication that needs uint128 or uint256 parsing.

contracts/contracts/CBOR.sol
CedarMist: We can't get rid of the error, Solidity has 256bit length prefixes for byte arrays.
I chose an arbitrary cutoff point (64kb), and have no made sure all paths are tested.
Beyond 64kb, user has other problems and is probably doing something wrong.

matevz: > Solidity has 256bit length prefixes for byte arrays.
> I chose an arbitrary cutoff point (64kb), and have no made sure all paths are tested.
> Beyond 64kb, user has other problems and is probably doing something wrong.

^^ This should go into the error description then:

```solidity
/// @notice Byte array is too long
/// @dev Solidity has 256bit length prefixes for byte arrays. We decided for
///      a reasonable cutoff point (64kb), so at most a 2-byte int describing
///      the array length. Beyond 64kb byte array, user will have other problems
///      and is probably doing something wrong.
error CBOR_Error_BytesTooLong();
```

contracts/contracts/CBOR.sol
matevz: Maybe put this into the error comment here.

```solidity
/// @notice The size of Uint exceeds 8 bytes
/// @dev Since the CBOR libraries for Javascript don't support encoding/decoding
///      of bigints beyond 64bit, we decided not to implement parsing those.
error CBOR_Error_UintTooLong();
```

contracts/contracts/Subcall.sol
```diff
@@ -98,7 +81,23 @@ library Subcall {
...
+        );
+
+        if (!success) {
+            revert SubcallError();
+        }
+
+        (status, data) = abi.decode(tmp, (uint64, bytes));
+    }
+
+    function subcall_static(string memory method, bytes memory body)
```
matevz, line 95:
```suggestion
    /**
     * @notice Submit a native message to the Oasis runtime layer using a static call.
     * Messages which re-enter the EVM module are forbidden: `evm.*`.
     * @param method Native message type.
     * @param body CBOR encoded body.
     * @return status Result of call.
     * @return data CBOR encoded result.
     */
    function subcall_static(string memory method, bytes memory body)
```

contracts/contracts/CBOR.sol
CedarMist: I have added support for uint128 and uint256.

It's not using tagged bignums (https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.3), I need to double-check with https://github.com/oasisprotocol/oasis-cbor for compatibility.

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,172 @@
...
+error CBOR_Error_InvalidUintSize(uint8);
+
+/// CBOR parsed value is out of expected range
+error CBOR_Error_ValueOutOfRange();
+
+error CBOR_Error_BytesTooLong(uint256 byteLength);
+
+function encodeUint(uint256 value) pure returns (bytes memory) {
+    // NOTE: we don't follow bignum tagged encoding
+    // See: https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.3
```
rube-de, line 27:
```suggestion
/// @dev we don't follow bignum tagged encoding
/// See: https://www.rfc-editor.org/rfc/rfc8949.html#section-3.4.3 
function encodeUint(uint256 value) pure returns (bytes memory) {
```

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,172 @@
+// SPDX-License-Identifier: Apache-2.0
+
+pragma solidity ^0.8.0;
+
+/// While parsing CBOR map, unexpected key
```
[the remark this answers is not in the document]
matevz: DYI natspec treats /// and /** comments as @notice if tag not specified: https://docs.soliditylang.org/en/latest/natspec-format.html#tags

So it's fine to omit it.

contracts/contracts/tests/TestCBOR.sol
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    function testUintRoundtrip(uint value)
+        external pure
+        returns (bool)
+    {
+        bytes memory encoded = CBOR.encodeUint(value);
+        (uint newOffset, uint result) = CBOR.parseUint(encoded, 0);
+        require( result == value, "value wrong!" );
+        require( newOffset == encoded.length, "didn't parse everything!" );
```
aefhm, line 37: (optional) I don't think coverage on this was there before, but maybe we want.

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,172 @@
...
+error CBOR_Error_InvalidKey();
+
+/// While parsing CBOR map, length is invalid, or other parse error
+error CBOR_Error_InvalidMap();
+
+/// While parsing CBOR structure, data length was unexpected
```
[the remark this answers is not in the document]
CedarMist: I've split this into `WrongMapSizeError` in `Subcall.sol` where it expects a map with a specific number of pairs, versus the CBOR error which denotes the map size is invalid (parse error).

contracts/contracts/CBOR.sol
```diff
@@ -0,0 +1,172 @@
...
+error CBOR_Error_InvalidLength(uint256);
+
+/// Value cannot be parsed as a uint
+error CBOR_Error_InvalidUintPrefix(uint8);
+
+/// Unsigned integer of unknown size
+error CBOR_Error_InvalidUintSize(uint8);
+
+/// CBOR parsed value is out of expected range
```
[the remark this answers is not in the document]
CedarMist: This would be a CBOR parse error, we encountered an unknown prefix while trying to parse a uint.

contracts/contracts/tests/TestCBOR.sol
CedarMist: That's a good point, need to ensure there are no out-of-bounds reads beyond buffer length.

Issue #337, opened 2024-07-19
sapphire-contracts: Flaky Gas padding test


matevz:
Gas padding in the CI tests is sometimes flaky for +/- 1. For example [here](https://github.com/oasisprotocol/sapphire-paratime/actions/runs/10006761516/job/27660080058?pr=332#step:16:130).

##### SUMMARY

```
  1) Gas Padding
       Gas Padding works as Expected:
      AssertionError: expected 132741 to equal 132742. The numerical values of the given "bigint" and "bigint" inputs were compared, and they differed.
      + expected - actual
      -132741
      +132742
      
      at Context.<anonymous> (test/gas.ts:26:40)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

PR #444, opened 2024-10-23
contracts/tests: Fix flaky SIWE and gas cost

### LINKED ISSUE #443: Fix SIWE intermittent test issue
Fails here: https://github.com/oasisprotocol/sapphire-paratime/blob/a4beb2acfcb993af66299022b5db2ba984908f8d/contracts/test/auth.ts#L167

If the block takes too much time to be mined, then the last block timestamp is still considered and the bearer is not expired yet. Fix the test so that it waits for the block to be mined with the timestamp greater than `expiration`.

matevz:
Fixes #443 by waiting for the block with sufficiently new timestamp instead of fixed time.
Another workaround for #337.

### DIFF
--- a/contracts/test/auth.ts
+++ b/contracts/test/auth.ts
@@ -160,10 +160,15 @@ describe('Auth', function () {
       siweStr4,
       await erc191sign(siweStr4, account),
     );
-    // Wait until the next second after expiration
-    await delay(expiration.getTime() - Date.now() + 1000);
-    const tx = await siweAuthTests.doNothing();
-    await tx.wait();
+    // Wait until the block time is greater than the expiration date
+    await new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


contracts/test/auth.ts
```diff
@@ -160,10 +160,15 @@ describe('Auth', function () {
       siweStr4,
       await erc191sign(siweStr4, account),
     );
-    // Wait until the next second after expiration
-    await delay(expiration.getTime() - Date.now() + 1000);
-    const tx = await siweAuthTests.doNothing();
-    await tx.wait();
+    // Wait until the block time is greater than the expiration date
+    await new Promise<void>((resolve, reject) => {
```
aefhm, line 164: (non-blocking) This might be a good helper/util function.

aefhm (approved):

CedarMist (approved):

PR #445, opened 2024-10-24
runtime: Bump Oasis SDK to 0.10.0 and prepare 0.9.0-testnet

matevz:


kostko (dismissed):

kostko:
We need to add explicit keys for mock SGX to support localnet builds. For `trusted_signers` you need to specify `TrustedSigners::unsafe_mock` under mock SGX.

CedarMist:
This works :D Thanks

```rust

    #[cfg(feature = "debug-mock-sgx")]
    fn trusted_signers() -> Option<TrustedSigners> {
        return Some(TrustedSigners::unsafe_mock());
    }

    #[cfg(not(feature = "debug-mock-sgx"))]
    fn trusted_signers() -> Option<TrustedSigners> {
        let tps = keymanager::trusted_policy_signers();
        // The `keymanager` crate may use a different version of `oasis_core`
        // so we need to convert the `TrustedSigners` between the versions.
        Some(TrustedSigners {
            signers: tps.signers.into_iter().map(|s| PublicKey(s.0)).collect(),
            threshold: tps.threshold,
        })
    }
```

runtime/src/lib.rs
```diff
@@ -142,7 +144,7 @@ impl sdk::Runtime for Runtime {
         if option_env!("OASIS_UNSAFE_SKIP_KM_POLICY") == Some("1")
             || cfg!(feature = "debug-mock-sgx")
         {
-            return Some(TrustedSigners::default());
+            return Some(TrustedSigners::unsafe_mock());
```
kostko, line 147: This won't work because `unsafe_mock` is only available under `debug-mock-sgx` feature. Use the variant mentioned by @CedarMist.

runtime/src/lib.rs
```diff
@@ -139,10 +141,9 @@ impl sdk::Runtime for Runtime {
 
     fn trusted_signers() -> Option<TrustedSigners> {
         #[allow(clippy::partialeq_to_none)]
-        if option_env!("OASIS_UNSAFE_SKIP_KM_POLICY") == Some("1")
-            || cfg!(feature = "debug-mock-sgx")
-        {
-            return Some(TrustedSigners::default());
+        #[cfg(feature = "debug-mock-sgx")]
+        if option_env!("OASIS_UNSAFE_SKIP_KM_POLICY") == Some("1") {
```
kostko, line 145: This also won't work, get rid of the env check, only the debug-mock-sgx feature is relevant in our localnet builds AFAIK.

runtime/Cargo.toml
```diff
@@ -21,8 +21,8 @@ debug = false
 keymanager = { git = "https://github.com/oasisprotocol/keymanager-paratime", tag = "v0.5.0" }
```
kostko, line 21:
```suggestion
keymanager = { git = "https://github.com/oasisprotocol/keymanager-paratime", tag = "v0.5.0-testnet" }
```

kostko (approved):

ptrus (approved):

PR #449, opened 2024-10-31
contracts: calldata encryption function (Solidity)

CedarMist:
Closes: #436

Implements calldata encryption to create encrypted transactions from contracts on Sapphire.

This also updates Cipher TS to accept an optional `nonce` parameter, this means we can do easier deterministic tests to verify the Solidity implementation matches the Typescript implementation.

### DIFF
--- a/clients/js/src/cipher.ts
+++ b/clients/js/src/cipher.ts
@@ -82,7 +82,10 @@ export abstract class Cipher {
   public abstract publicKey: Uint8Array;
   public abstract epoch?: number;
 
-  public abstract encrypt(plaintext: Uint8Array): {
+  public abstract encrypt(
+    plaintext: Uint8Array,
+    nonce?: Uint8Array,
+  ): {
     ciphertext: Uint8Array;
     nonce: Uint8Array;
   };
@@ -93,7 +96,10 @@ export abstract class Cipher {
   ): Uint8Array;
 
   /** Encrypts the plaintext and encodes it for sending. */
-  public encryptCall(calldata?: BytesLike | null): BytesLike {
+  public encryptCall(
+    calldata?: BytesLike | null,
+    nonce?: Uint8Array,
+  ): BytesLike {
     // Txs without data are just balance transfers, and all data in those is public.
     if (calldata === undefined || calldata === null || calldata.length === 0)
       return '';
@@ -104,7 +110,8 @@ export abstract class Cipher {
 
     const innerEnvelope = cborEncode({ body: getBytes(calldata) });
 
-    const { ciphertext, nonce } = this.encrypt(innerEnvelope);
+    let ciphertext: Uint8Array;
+    ({ ciphertext, nonce } = this.encrypt(innerEnvelope, nonce));
 
     const envelope: Envelope = {
       format: this.kind,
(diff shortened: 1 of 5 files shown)


rube-de:
Running the test against localnet fails for me
```
  1) CalldataEncryption
       testEncryptCallData:

      AssertionError: expected '0xa264626f6479a462706b58204a0269de680…' to equal '0xa264626f6479a462706b58204a0269de680…'
      + expected - actual

      -0xa264626f6479a462706b58204a0269de680edc05553943543d70c5df9e23f6af53381eded9641b5d08df7b2d646461746158184bf966b9e87a715c26b2428996c85589e1ba7e2c060487206565706f636800656e6f6e63654fb9cb3f82f8a608902dcb321ae611a766666f726d617401
      +0xa264626f6479a462706b58204a0269de680edc05553943543d70c5df9e23f6af53381eded9641b5d08df7b2d64646174615818cbca0d4444a619c839a845d2b045ccf4437a861047e59b5f6565706f636800656e6f6e63654f0d0922c9861097bab82a5a77b15d8666666f726d617401
      
       at Context.<anonymous> (test/calldata.spec.ts:50:22)
```

CedarMist:
That could happen if you don't build the clients/js package, the solidity tests require repeatability so I had to modify the client to pass a custom nonce.

If the clients/js package isn't rebuilt, the solidity tests will use the last build you did which will generate a nonce every time and ignore the one passed in.

clients/js/src/cipher.ts
```diff
@@ -251,11 +258,14 @@ export class X25519DeoxysII extends Cipher {
     this.cipher = new deoxysii.AEAD(new Uint8Array(this.key)); // deoxysii owns the input
   }
 
-  public encrypt(plaintext: Uint8Array): {
+  public encrypt(
+    plaintext: Uint8Array,
+    nonce?: Uint8Array,
```
rube-de, line 263:
```suggestion
    nonce: Uint8Array = randomBytes(deoxysii.NonceSize),
```
Instead of check and assign, it could be defaulted to randomBytes directly

clients/js/src/cipher.ts
```diff
@@ -251,11 +258,14 @@ export class X25519DeoxysII extends Cipher {
...
+  ): {
     ciphertext: Uint8Array;
     nonce: Uint8Array;
   } {
-    const nonce = randomBytes(deoxysii.NonceSize);
+    nonce = nonce ?? randomBytes(deoxysii.NonceSize);
```
rube-de, line 268:
```suggestion
```
if default is set, this can be removed

contracts/test/calldata.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  it('roundtrip encryption', async () => {
+    // Tests must be submitted from an account which has a balance
+    // But can't get access to the signer private key from here
+    // So, assume the 0xf39F address is being used to run tests
+    const myAddr = '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266';
+    const myKey =
+      '0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80';
+    const [signerAddr] = await ethers.getSigners();
```
rube-de, line 61: Initialisation could be moved in the `before` block, for better structure and if more tests are added in future that need this values too.

rube-de (approved):

PR #457, opened 2024-11-19
docs: Update Draw.io SVG to dark mode

aefhm:
## Description

Supports https://github.com/oasisprotocol/docs/issues/308


docs/images/gasless-on-chain-signer.svg
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Do not edit this file with editors other than draw.io -->
```
aefhm, line 2: @matevz lost your earlier comment, but fixed the export type. New format seems way bigger unfortunately.

matevz: There seems to be a new PNG image embedded as base64 inside the svg file? Can you investigate?

aefhm: This appears to be a result of embedding the font in the SVG. The other SVGs had this embedding [already](https://github.com/oasisprotocol/oasis-sdk/pull/2068/files#diff-0f4aed62c4be4fb640fce809c969fe0f8205a9f6057f544713d5d51552886206).

We could otherwise choose to drop the font. I can't imagine Helvetica not on a machine.

matevz: Interesting. Yeah, we should avoid embedding the fonts and simply use the web safe ones.

aefhm: Lemme fix the other ones then. We historically had the fonts there.

matevz (approved):
Remove the embedded fonts, otherwise lgtm.

PR #471, opened 2024-12-12
ci: Fix Docker options after removing oasis-deposit

### LINKED ISSUE #465: Update account funding parameters on CI
We need to update lines like these:
https://github.com/oasisprotocol/sapphire-paratime/blob/main/.github/workflows/ci-test-go.yaml#L29

`oasis-deposit` has been removed from the container, so we need to pass the options directly to the startup script.

abukosek:
Note that I haven't found a way to pass the script arguments to the end of the Docker run command from GitHub Actions, but the nice part is that they don't really matter, since the test mnemonic and 5 accounts is already the default :man_shrugging:

Closes #465.

### DIFF
--- a/.github/workflows/ci-playwright.yaml
+++ b/.github/workflows/ci-playwright.yaml
@@ -16,7 +16,7 @@ jobs:
           - 8545:8545
           - 8546:8546
         env:
-          OASIS_DEPOSIT_BINARY: /oasis-deposit -test-mnemonic -n 5
+          OASIS_DOCKER_START_EXPLORER: no
         options: >-
           --rm
           --health-cmd="test -f /CONTAINER_READY"
--- a/.github/workflows/ci-test-go.yaml
+++ b/.github/workflows/ci-test-go.yaml
@@ -26,7 +26,7 @@ jobs:
           - 8545:8545
           - 8546:8546
         env:
-          OASIS_DEPOSIT_BINARY: /oasis-deposit -test-mnemonic -n 2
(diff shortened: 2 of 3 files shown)


matevz:
Can you pass the `-to` and `-n` inside the `options:` section? Or did you just hit https://github.com/actions/runner/pull/1152 ?

matevz (approved):

abukosek:
> Can you pass the `-to` and `-n` inside the `options:` section? Or did you just hit [actions/runner#1152](https://github.com/actions/runner/pull/1152) ?

Yeah, I tried both `options` and `args` and neither worked, because in both cases the options are treated as arguments to the `docker run` command instead of as arguments to the script inside the container :(

abukosek:
> > > Can you pass the `-to` and `-n` inside the `options:` section? Or did you just hit [actions/runner#1152](https://github.com/actions/runner/pull/1152) ?
> > 
> > 
> > Yeah, I tried both `options` and `args` and neither worked, because in both cases the options are treated as arguments to the `docker run` command instead of as arguments to the script inside the container :(
> 
> What about overriding `--entrypoint '/start.sh -to 0x...'` ?

That is super-hacky and harder to maintain. In the case of our CI, funding 2 accounts vs. 5 doesn't result in a noticeable speed improvement to warrant the extra maintenance burden IMO :)

abukosek:
> It is for cases when a specific dApp CI tests require non-default accounts to be funded, or none at all.

Yeah, in that case overriding the entrypoint is an option.

PR #477, opened 2024-12-23
Add encryptCallData() to gasless chapter and example

### LINKED ISSUE #463: docs: Use new encryptCallData in EIP155 gasless chapter

matevz:
This PR:
- documents `encryptCallData()` in `gasless.md` to generate encrypted gasless transaction
- adds `encryptCallData()` to `examples/onchain-signer` and tests it in combination with the wrapped/unwrapped provider and encrypted/plain gasless tx.
- removes some unneeded sleeps from the onchain-signer tests since the recent sapphire-localnet healthcheck correctly waits that the EVM is spun up before running them

[PREVIEW](https://deploy-preview-477--oasisprotocol-sapphire-paratime.netlify.app/dapp/sapphire/gasless)

Fixes #463.

### DIFF
--- a/examples/onchain-signer/hardhat.config.ts
+++ b/examples/onchain-signer/hardhat.config.ts
@@ -26,7 +26,17 @@ const config: HardhatUserConfig = {
     'sapphire-testnet': { ...sapphireTestnet, accounts },
     'sapphire-localnet': { ...sapphireLocalnet, accounts },
   },
-  solidity: '0.8.20',
+  solidity: {
+    version: '0.8.20',
+    settings: {
+      // XXX: Needs to match https://github.com/oasisprotocol/sapphire-paratime/blob/main/contracts/hardhat.config.ts
+      optimizer: {
+        enabled: true,
+        runs: (1 << 32) - 1,
+      },
+      viaIR: true,
+    },
+  },
 };
 
 export default config;
--- a/examples/onchain-signer/test/CommentBox.ts
+++ b/examples/onchain-signer/test/CommentBox.ts
@@ -3,6 +3,10 @@ import { ethers } from 'hardhat';
 import { parseEther, Wallet } from 'ethers';
 import { CommentBox, Gasless } from '../typechain-types';
 import { EthereumKeypairStruct } from '../typechain-types/contracts/Gasless';
+import {
+  isCalldataEnveloped,
+  wrapEthereumProvider,
+} from '@oasisprotocol/sapphire-paratime';
 
 describe('CommentBox', function () {
   let commentBox: CommentBox;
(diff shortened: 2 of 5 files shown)


examples/onchain-signer/hardhat.config.ts
```diff
@@ -26,7 +26,17 @@ const config: HardhatUserConfig = {
     'sapphire-testnet': { ...sapphireTestnet, accounts },
     'sapphire-localnet': { ...sapphireLocalnet, accounts },
   },
-  solidity: '0.8.20',
+  solidity: {
+    version: '0.8.20',
+    settings: {
+      // XXX: Needs to match https://github.com/oasisprotocol/sapphire-paratime/blob/main/contracts/hardhat.config.ts
```
ptrus, line 32: Based on some of my quick testing I believe it is not about matching, but actually things fail without this optimization enabled.

For example, removing this optimization in contracts/hardhat.config.ts makes the tests in contracts/ fail as well.

Do we know why this happens? Should someone have investigate this further?

examples/onchain-signer/test/CommentBox.ts
```diff
@@ -33,55 +34,60 @@ describe('CommentBox', function () {
...
+        innercall,
+      );
+    } else {
+      tx = await gasless.makeProxyTx(await commentBox.getAddress(), innercall);
+    }
 
-    const tx = await commentBox.comment('Hello, world!');
-    await tx.wait();
+    // TODO: https://github.com/oasisprotocol/sapphire-paratime/issues/179
+    const response = await provider.broadcastTransaction(tx);
```
rube-de, line 56: calldata encryption could be checked here (requires sapphire-paratime import)
```suggestion
    const response = await provider.broadcastTransaction(tx);
    expect(isCalldataEnveloped(response.data)).eq(!plain);
```

also is the  TODO still open?
I used `await wrapEthereumProvider(provider).broadcastTransaction(tx)` and didn't got an error.

examples/onchain-signer/hardhat.config.ts
matevz: This would require investigation. I opened https://github.com/oasisprotocol/sapphire-paratime/issues/479 so we don't forget it.

examples/onchain-signer/test/CommentBox.ts
matevz: > I used `await wrapEthereumProvider(provider).broadcastTransaction(tx)` and didn't got an error.

https://github.com/oasisprotocol/sapphire-paratime/issues/179 is not relevant anymore. Fixed.

matevz: > calldata encryption could be checked here (requires sapphire-paratime import)

Good idea!

rube-de (approved):

ptrus (approved):

PR #483, opened 2025-01-14
Fix Avalanche fuji chain ID reference

aefhm:
## Description

Fix incorrect chain [ID](https://docs.avax.network/dapps/chain-settings).


rube-de (approved):

aefhm:
We should bump package versioning, but probably not in a rush.

Issue #511, opened 2025-02-18
Update docs link to correct SIWE repo


aefhm:
https://docs.oasis.io/build/sapphire/examples#siwe should link to `demo-starter` at `master`

PR #517, opened 2025-02-18
docs: Fix wording and link

### LINKED ISSUE #515: Polish view call authentication docs
## Description

We want to show that signed queries only work in Go & Python client, and inform devs that this for UX reasons because JS client signed queries aren't available.

aefhm:
## Description

Close #515 and #511.

### DIFF
--- a/docs/develop/authentication.md
+++ b/docs/develop/authentication.md
@@ -51,8 +51,8 @@ In Sapphire, you need to consider the following types of contract calls:
     [EIP-712] defines a format for signing view calls with the keypair of your
     Ethereum account. Sapphire will validate such signatures and automatically
     set the `msg.sender` parameter in your contract to the address of the
-    signing account. This method is mostly appropriate for backend services,
-    since the frontend would require user interaction each time.
+    signing account. This method is most appropriate for backend services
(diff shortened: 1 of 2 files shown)


docs/develop/authentication.md
```diff
@@ -51,8 +51,8 @@ In Sapphire, you need to consider the following types of contract calls:
     [EIP-712] defines a format for signing view calls with the keypair of your
     Ethereum account. Sapphire will validate such signatures and automatically
     set the `msg.sender` parameter in your contract to the address of the
-    signing account. This method is mostly appropriate for backend services,
-    since the frontend would require user interaction each time.
+    signing account. This method is most appropriate for backend services
+    because frontend/JS applications would require user interaction each time.
```
rube-de, line 55: I wouldn't add the JS mention; in the end, it's a case of whether the code has access to the private key and can sign the message, or the message has to be signed by a second party (which is mostly the case for webdapps).

You can have JS applications which take/generate your private key, and then they have it afterwards to sign stuff on your behalf.

aefhm: Makes sense. Removing "JS"

rube-de (approved):

PR #520, opened 2025-02-24
docs: Include info for signed view call change

### LINKED ISSUE #516: Add notice on V1 client and MetaMask
## Description

Add notice that the V1 client had signed view calls, but this does not work well with MetaMask.

aefhm:
## Description

Close #516 which was a note of the planned release, but I think this is more of a nice to have now.

### DIFF
--- a/docs/develop/authentication.md
+++ b/docs/develop/authentication.md
@@ -246,6 +246,16 @@ your Web3 endpoint.
 
 ### via signed queries
 
+:::info
+
+Sapphire TypeScript wrapper v1 included signed view calls which resulted in
+MetaMask popups for every view call. We recommend authenticating [via SIWE token]
+with the Sapphire TypeScript wrapper v2 going forward.
+
+:::
+
+[via SIWE token]: #via-siwe-token
+
 [EIP-712] proposed a method to show data to the user in a structured fashion so
 they can verify it before signing. In the browser however, apps requiring signed
 view calls would trigger user interaction with their wallet each time—sometimes


docs/develop/authentication.md
```diff
@@ -256,6 +256,14 @@ interaction. This is possible because a backend service connects to a trusted
 site and executes trusted code, so it's fine to sign the necessary view calls
 non interactively.
 
+:::info
+
+Sapphire TypeScript wrapper v1 included signed view calls which resulted in
+multiple MetaMask messages. We recommend the usage of SIWE tokens going
+forward with v2.
```
rube-de, line 263:
```suggestion
Sapphire TypeScript wrapper v1 included signed view calls which resulted in
MetaMask popups for every view call. We recommend authentication [via SIWE token]
going forward with the Sapphire TypeScript wrapper v2
```

I think on the line of 'resulted in MetaMask popups for every view call' is clearer.
In the 2nd sentence maybe a bit more explicit that it's still about typescript and link to the section.

rube-de: I was also contemplating whether the infobox is in the right position, or if we should provide a recommendation earlier.

matevz (commented):
Let's not clutter current doc with obsolete client versions info. If you really think this is necessary, I would reword the first paragraph and insert a short sentence about v1 TypeScript Sapphire wrapper.

aefhm:
> Let's not clutter current doc with obsolete client versions info. If you really think this is necessary, I would reword the first paragraph and insert a short sentence about v1 TypeScript Sapphire wrapper.

Hmm. I really don't think confusion from existing users will be a thing in a migration, so yes, reducing information should be more clear going forward.

PR #550, opened 2025-03-26
sapphire-wagmi-v2: Add EIP-6963 support

### LINKED ISSUE #499: sapphire-wagmi-v2: Add EIP-6963 support
##### SUMMARY
Current implementation of `@oasisprotocol/sapphire-wagmi-v2` only supports default injected EIP-1193 provider:

> Please note that while [EIP-6963](https://eips.ethereum.org/EIPS/eip-6963) (Multi Injected Provider Discovery) is supported by Wagmi it is only possible to wrap the default injected [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) compatible provider. For this reason you must disable MIPD support in the Wagmi configuration as additional discovered providers will not be Sapphire wrapped.

Currently setting to:
```js
   {
      // ...
      multiInjectedProviderDiscovery: false,
      // ...
   }
```

Disables EIP-6963 functionality. 

##### ISSUE TYPE
- Feature request

##### ADDITIONAL INFORMATION

Currently it is easy to misuse `@oasisprotocol/sapphire-wagmi-v2`. If you don't use EIP-1193 injected provider, and use one provided by EIP-6963, could result in provider not being correctly wrapped. Implementing EIP-6963 would result in clearer way of how to use `@oasisprotocol/sapphire-wagmi-v2` and package not being misused.

lubej:
Closes #499


integrations/wagmi-v2/src/constants.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * @license Apache-2.0
+ */
+
+import type { Rdns } from "./types.js";
+
+export const SUPPORTED_RDNS: Rdns[] = [
+	"io.metamask",
+	"io.metamask.mobile",
+	"com.brave.wallet",
```
lubej, line 10: Wrapping some providers might result in breaking the execution, when I was testing it was failing with OKX Wallet. Thats why the whitelist was added.

integrations/wagmi-v2/src/index.ts
```diff
@@ -63,3 +84,141 @@ export function injectedWithSapphire(
...
+ *
+ *    export const config = createSapphireConfig({
+ *    	sapphireConfig: {
+ *          replaceProviders: true
+ *    	}
+ *      ...
+ *    });
+ * ```
+ *
+ * In case your dApp is multichain, the bellow example will create duplicate connectors for Sapphire:
```
aefhm, line 104:
```suggestion
 * In case your dApp is multichain, the below example will create duplicate connectors for Sapphire:
```

integrations/wagmi-v2/README.md
```diff
@@ -18,67 +18,111 @@ and gas estimations.
...
 
-Next, in your Wagmi config definition, setup Sapphire wrapping for the injected
-provide using `injectedWithSapphire()` and then define the transports using the
-`sapphireHttpTransport()` function.
+### [EIP-6963] Multi Injected Provider Discovery
+
+#### Single chain - Sapphire
+
+When your application connects exclusively to Sapphire networks, the most 
+straightforward approach is to set `sapphireConfig.replaceProviders = true`.
```
aefhm, line 34: Default is `false` right?

integrations/wagmi-v2/README.md
```diff
@@ -18,67 +18,111 @@ and gas estimations.
...
+
 [EIP-6963]: https://eips.ethereum.org/EIPS/eip-6963
-[EIP-1193]: https://eips.ethereum.org/EIPS/eip-1193
 
-To connect to your `sapphire-localnet` instance, define a custom chain:
+### [EIP-1193] Injected provider
+
+In your Wagmi config definition, setup Sapphire wrapping for the injected
+provide using `injectedWithSapphire()` and then define the transports using the
+`sapphireHttpTransport()` function.
```
aefhm, line 101:
```suggestion
In your Wagmi config definition, wrap the injected
provider for Sapphire using `injectedWithSapphire()`, then define the transports using the
`sapphireHttpTransport()` function.
```

integrations/wagmi-v2/src/index.ts
```diff
@@ -63,3 +84,141 @@ export function injectedWithSapphire(
...
+										},
+									);
+								}
+							} else if (
+								wrappedProvidersFilter(
+									announceProviderEvent.detail.info.rdns,
+								) &&
+								!isWrappedEthereumProvider(
+									announceProviderEvent.detail.provider,
+								)
```
aefhm, line 185: (optional) Maybe name and lift this up earlier 🤔

lubej: Like so https://github.com/oasisprotocol/sapphire-paratime/pull/550/commits/720f2524f8ee4bfac6ead7164cea26a8c00fb256?

examples/wagmi-v2/test/e2e.spec.ts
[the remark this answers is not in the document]
lubej: Right, removed leading spaces.

integrations/wagmi-v2/src/constants.ts
matevz: Is incompatibility with OKX a known bug?

integrations/wagmi-v2/src/index.ts
aefhm: Works for me :)

I was thinking just something basic would do too.

```ts
condition = wrappedProvidersFilter(announceProviderEvent.detail.info.rdns,) && !isWrappedEthereumProvider(announceProviderEvent.detail.provider)
```

integrations/wagmi-v2/src/constants.ts
lubej: image
Looks like the OKX wallet does not support Sapphire network.

integrations/wagmi-v2/src/index.ts
lubej: I am not in favor of extracting if statement/s condition to a variable, it decreases readability IMO. Maybe in cases where it would speed up execution.

lubej:
Playwright test seems to be failing, but I could only reproduce it once. Needs investigation, seems there is an alert that need to be dismissed upon deploying the contract.

matevz:
> Playwright test seems to be failing, but I could only reproduce it once. Needs investigation, seems there is an alert that need to be dismissed upon deploying the contract.

Playwright seems flaky. Also failed on @ZigaMr PR https://github.com/oasisprotocol/sapphire-paratime/actions/runs/14178490717/job/39718975456?pr=552

integrations/wagmi-v2/README.md
```diff
@@ -18,67 +18,111 @@ and gas estimations.
...
-supported by Wagmi it is only possible to wrap the default injected [EIP-1193]
-compatible provider. For this reason you must disable MIPD support in the
-Wagmi configuration as additional discovered providers will not be Sapphire
-wrapped.
+#### Multichain
+
+For multichain applications, you need to carefully manage which providers use 
+end-to-end encryption. It's essential to avoid using encrypted connectors for 
+non-Sapphire chains, and conversely, to ensure Sapphire chains use the 
+appropriate encrypted connectors.
```
aefhm, line 64:
```suggestion
For multichain applications, you must manage which providers use 
end-to-end encryption. It's essential to avoid using encrypted connectors for 
non-Sapphire chains, and conversely, to ensure Sapphire chains use the 
appropriate encrypted connectors.
```

aefhm (approved):
Interface seems reasonable to me.

Not introduced in this PR, but I find it weird that `sapphire-wagmi-v2` whitespace is so different from `sapphire-paratime`, and prefer the denser/smaller one.

integrations/wagmi-v2/README.md
```diff
@@ -18,67 +18,111 @@ and gas estimations.
...
+`sapphire.${rdns}` for each selected provider. For example, in an application 
+supporting both Ethereum and Sapphire networks, you would use the 
+`sapphire.io.metamask` connector specifically for Sapphire chains and the 
+standard `io.metamask` connector for Ethereum chain.
+
 [EIP-6963]: https://eips.ethereum.org/EIPS/eip-6963
-[EIP-1193]: https://eips.ethereum.org/EIPS/eip-1193
 
-To connect to your `sapphire-localnet` instance, define a custom chain:
+### [EIP-1193] Injected provider
```
rube-de, line 97: Question for understanding:
The above createSapphireConfig() covers already injected provider, is there any reason to not use it?

lubej: `createSapphireConfig` does not cover Injected provider. So the way this works is, that it is a race condition which wallet will "register" itself to `window.ethereum`. But each of the wallets will announce themselves as EIP-6963. So there really is no reason to use both. Potentially you could use both, Injected provider with `createSapphireConfig`. One potential use for Injected provider could be, that you don't want to implement the "wallet picker/EIP-6963 selection" modal, and you just use Injected provider. And some of the wallets, have a mode to "bypass" it to next registered wallet. Or the wallet does not implement EIP-6963, and just uses `window.ethereum` registration. I still think we need a way to cover both use cases, even though preferred mode of usage is EIP-6963.

integrations/wagmi-v2/src/index.ts
```diff
@@ -63,3 +84,131 @@ export function injectedWithSapphire(
...
+			type: string,
+			callback: EventListenerOrEventListenerObject | null,
+			options?: AddEventListenerOptions | boolean,
+		): void {
+			if (type === EIP6963_ANNOUNCE_PROVIDER_EVENT_NAME) {
+				_addEventListener.call(
+					this,
+					type,
+					(event: Event) => {
+						let patchCustomEvent = null;
```
lukaw3d, line 150: This looks like it breaks removeEventListener (callback function won't match inline patch function)

lukaw3d: If removing EIP6963_ANNOUNCE_PROVIDER_EVENT_NAME listeners isn't common then I would just add a headsup comment

lubej: I haven't seen a codebase, that would try to remove the event. But yeah, I see the potential issue. Thats why I didn't add "restore" to original `_addEventListener`, which could cause memory leaks, because I did not expect that someone would try to remove the event listener. 

Either way, I added the comment here: https://github.com/oasisprotocol/sapphire-paratime/pull/565. I hope that is enough for now, and if there is an use-case for it. I am happy to implement more robust solution, that would "reverse" event patching.

PR #555, opened 2025-03-31
clients/js: Make NETWORKS compatible with wallet_addEthereumChain

matevz:
We have the "Add Sapphire network to Metamask" in the growing number of places (docs, faucet, all the dApps). Since most of those apps explicitly or implicitly import `@oasisprotocol/sapphire-paratime`, let's unify the network details in a single place.

For now, I made it compatible with https://docs.metamask.io/wallet/reference/json-rpc-methods/wallet_addethereumchain/ so one can directly use it in the UI.

But there seems to be some other competing formats:
- ethers Network object: https://docs.ethers.org/v6/api/providers/#Network
- ethereum chains https://github.com/ethereum-lists/chains/blob/master/tools/schema/chainSchema.json

### DIFF
--- a/clients/js/src/networks.ts
+++ b/clients/js/src/networks.ts
@@ -22,22 +22,52 @@ export const SAPPHIRE_LOCALNET_HTTP_PROXY_HOST = globalThis.process?.env
   : 'localhost';
 
 const localnetParams = {
+  chainName: 'Oasis Sapphire Localnet',
   chainId: 0x5afd,
   defaultGateway: `http://${SAPPHIRE_LOCALNET_HTTP_PROXY_HOST}:${SAPPHIRE_LOCALNET_HTTP_PROXY_PORT}`,
+  rpcUrls: [
+    `http://${SAPPHIRE_LOCALNET_HTTP_PROXY_HOST}:${SAPPHIRE_LOCALNET_HTTP_PROXY_PORT}`,
+  ],
+  nativeCurrency: {
+    name: 'Oasis TEST',
+    symbol: 'TEST',
+    decimals: 18,
+  },
+  blockExplorerUrls: ['http://localhost:8548/localnet/sapphire'],
   runtimeId:
     '0x8000000000000000000000000000000000000000000000000000000000000000',
 };
 
 const mainnetParams = {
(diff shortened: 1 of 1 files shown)


clients/js/src/networks.ts
```diff
@@ -22,22 +22,49 @@ export const SAPPHIRE_LOCALNET_HTTP_PROXY_HOST = globalThis.process?.env
...
+    decimals: 18,
+  },
+  blockExplorerUrls: ['http://localhost:8548/localnet/sapphire'],
+  iconUrls: ['https://docs.oasis.io/img/favicon.svg'], // TODO
   runtimeId:
     '0x8000000000000000000000000000000000000000000000000000000000000000',
 };
 
 const mainnetParams = {
+  chainName: "Oasis Sapphire",
```
rube-de, line 41: For consistency, we could name this Oasis Sapphire Mainnet, as the parameter also suggests and we have in our Network documentation also mentioned Mainnet.
Ethereum is also listed as Ethereum Mainnet on https://github.com/ethereum-lists/chains/ (Sapphire should then also be renamed here to Oasis Sapphire Mainnet).

matevz: We decided a while a go to drop Mainnet where possible, because there could be multiple Testnets in the future, but only on Mainnet. But now I see more and more chains are bringing this term back, for example in the Metamask dropdown.

So I agree, perhaps it's time to re-add it back. Similar for Emerald.

matevz:
This PR needs decision on official icons for Sapphire Mainnet/Testnet. (also, do we separate between Emerald/Sapphire?) 

Related https://github.com/MetaMask/metamask-extension/pull/31588

lukaw3d:
I would unblock this by copying icon 6 times to assets.oasis.io. When decision comes, update it there :shrug:

lubej (approved):

clients/js/src/networks.ts
matevz: On a second thought I wouldn't prefer long names such as "Oasis Sapphire Mainnet". The shorter the better IMO, so a simple "Oasis Sapphire" in production looks more sleek.

PR #565, opened 2025-04-09
sapphire-wagmi-v2: Fix removeEventListener bug

lubej:


integrations/wagmi-v2/src/index.ts
```diff
@@ -116,6 +116,9 @@ export function injectedWithSapphire(
  * });
  * ```
  *
+ * In order to correctly remove eip6963:announceProvider event. Please use signal approach:
```
matevz, line 119: If this is the preferred flow, then put it inside the code block above?

integrations/wagmi-v2/src/index.ts
```diff
@@ -135,79 +135,108 @@ export function createSapphireConfig<
...
+				const wrapperFn = listenerMap.get(callback);
+
+				listenerMap.delete(callback);
+
+				_removeEventListener.call(this, type, wrapperFn ?? null, options);
+			} else {
+				_removeEventListener.call(this, type, callback, options);
+			}
+		},
+	});
```
lukaw3d, line 238: This seems annoying to maintain.

Can we reimplement the whole thing, and instead of patching
`addEventListener if eip6963:announceProvider wrapEthereumProvider`
could patch
`window.dispatchEvent if eip6963:announceProvider wrapEthereumProvider`

lubej:
Closing in favor of https://github.com/oasisprotocol/sapphire-paratime/pull/577

PR #574, opened 2025-04-18
docs: improve contextOrHash parameter documentation

rube-de:
Enhance the documentation for the `sign()` and `verify()` functions in Sapphire.sol to clarify how the `contextOrHash` parameter should be used with different signing algorithms:

- Correct the expected hash size for Secp384r1PrehashedSha384 from 32 bytes to 48 bytes
- Add detailed explanations of parameter behavior for each algorithm type
- Explicitly document that Sr25519 signatures MUST use "substrate" as the context for compatibility with Substrate off-chain signing using { withType: true }
- Add clear code examples showing proper usage for different algorithms
- Maintain consistent documentation between sign() and verify() functions

### DIFF
--- a/contracts/contracts/Sapphire.sol
+++ b/contracts/contracts/Sapphire.sol
@@ -359,15 +359,30 @@ library Sapphire {
 
     /**
      * @notice Sign a message within the provided context using the specified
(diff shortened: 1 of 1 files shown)


matevz (approved):
Good job, thanks!

PR #596, opened 2025-06-10
Try to fail foundry due to wrong EVM version

lukaw3d:
https://github.com/oasisprotocol/sapphire-paratime/blob/052e6df2acc6fa1e08529b4f79900a4918b9f4a4/examples/foundry/foundry.toml#L1-L7 

doesn't specify evm version (https://github.com/oasisprotocol/demo-rofl-chatbot/pull/17), and

https://github.com/oasisprotocol/sapphire-paratime/blob/052e6df2acc6fa1e08529b4f79900a4918b9f4a4/.github/workflows/contracts-test.yaml#L66

seems to use a latest stable version of foundry, so this should fail

### DIFF
--- a/examples/foundry/src/Counter.sol
+++ b/examples/foundry/src/Counter.sol
@@ -17,4 +17,8 @@ contract Counter is SapphireDecryptor {
     function withdraw() external {
         payable(msg.sender).transfer(address(this).balance);
     }
+
+    function thisCompilesToPush0AndShouldDetectWrongEvmVersion() public pure returns (bool) {
+        return false; // This will compile to PUSH0 in new unsupported EVM versions
+    }
 }


examples/foundry/src/Counter.sol
```diff
@@ -17,4 +17,8 @@ contract Counter is SapphireDecryptor {
     function withdraw() external {
         payable(msg.sender).transfer(address(this).balance);
     }
+
+    function thisCompilesToPush0AndShouldDetectWrongEvmVersion() public pure returns (bool) {
+        return false; // This will compile to PUSH0 in new unsupported EVM versions
+    }
```
lukaw3d, line 23: nevermind :shrug: 

This deploys to Sapphire Testnet successfully
```
pragma solidity ^0.8.13;

contract A {
    function thisCompilesToPush0AndShouldDetectWrongEvmVersion() public pure returns (bool) {
        return false; // This will compile to PUSH0 in new unsupported EVM versions
    }
}
```

PR #601, opened 2025-07-16
fix: account switching an dappwright

rube-de:
Metamask seems to name the accounts weird in the setup (maybe different every version?)
"jvh" corresponds to address 0x70997970C51812dc3A010C7d01b50e0d17dc79C8

Screenshot 2025-07-16 at 21 33 39


lukaw3d (approved):

lubej (approved):
Yeah, not sure either where the naming comes from.

PR #607, opened 2025-08-04
wagmi-v2: Wrap connector

lubej:
**BREAKING CHANGE**

With https://github.com/oasisprotocol/sapphire-paratime/pull/550 there were a couple of problems that revealed during usage. Even though it successfully wrapped EIP-6963 providers, the implementation did not work for WalletConnect and MetaMask "in-app" communication between mobile Chrome and mobile MetaMask for example.

This PR addresses the above issues and resolves them by changing the underlying API, which now wraps the connectors directly. By doing so, all the issues with previous implementation were successfully resolved.


matevz (commented):
I tested the example on mobile and for Metamask mode while switching between the browser and the Metamask wallet back and forth, the contract deployment tx and its address is lost somewhere along the way. Can you take a look?

Wallet connect mode worked fine for me on mobile Chrome, but not Firefox.

On desktop, Metamask mode worked. And Wallet connect does not know how to communicate with my Metamask browser extension, right?

integrations/wagmi-v2/src/index.ts
```diff
@@ -85,124 +68,95 @@ export function injectedWithSapphire(
...
+ * import { wrapConnectorWithSapphire } from '@oasisprotocol/sapphire-wagmi-v2';
+ * import { walletConnect } from '@wagmi/connectors';
  *
- * export const config = createSapphireConfig({
- *   sapphireConfig: {
- *     replaceProviders: false,
- *     // Define which providers you want to wrap via RDNS
- *     wrappedProvidersFilter: (rdns) => ['io.metamask'].includes(rdns)
- *   }
+ * export const config = createConfig({
```
matevz, line 87: Where is createConfig() defined?

integrations/wagmi-v2/src/index.ts
```diff
@@ -85,124 +68,95 @@ export function injectedWithSapphire(
...
+ *   connectors: [
+ *     wrapConnectorWithSapphire(
+ *       walletConnect({ projectId: 'your-project-id' }),
```
matevz, line 90: How do you wrap plain wagmi connector? e.g. If you don't want to use wallet connect / rainbowkit?

matevz (commented):
Can you also update https://github.com/oasisprotocol/sapphire-paratime/blob/ml/wagmi-v2-wrap-connector/integrations/wagmi-v2/README.md for usage?

integrations/wagmi-v2/src/index.ts
lubej: With `wrapConnectorWithSapphire` utility function.

integrations/wagmi-v2/src/index.ts
lubej: `import { createConfig } from "wagmi";` comes from wagmi.

matevz:
Examples should use Sapphire Testnet by default. It seems mine jumped to the Sapphire Mainnet and I can't seem to switch it to another network in the app. Changing it in Metamask works, but the "Deploy contract" still wants to use Mainnet.

image

matevz (commented):
Can you check three more things:
1. On desktop, if I select Wallet connect, I can't seem to copy the QR code with the button on the right (it just says "Failed to copy").
2. Once connected, if I change the network in Metamask, the change does not appear in the dApp. Can you add a hook?
3. The connect wallet button should switch to Sapphire Testnet or suggest to add it. This should already be part of the example IMO.

matevz (approved):
Ok, let's get this in and we'll do followup PRs.

lubej:
> > Is this something we should release?
> 
> Yes. #656

@matevz ideally we would include this https://github.com/oasisprotocol/sapphire-paratime/pull/634 as well for the next release.

PR #608, opened 2025-08-05
Add new styling to mermaid diagrams

kaja-osojnik:
Part II of [#1378](https://github.com/oasisprotocol/docs/pull/1378)


matevz (approved):
These are really really nice. Thank you!

PR #609, opened 2025-08-07
Generate mermaid diagrams on CI with minor differences

lukaw3d:
CI in oasisprotocol/docs https://github.com/oasisprotocol/docs/pull/1378 complains that generated diagrams don't match those generated locally in https://github.com/oasisprotocol/sapphire-paratime/pull/608

so these are copied from CI logs


lukaw3d:
You prefer transparent like this?
image

PR #616, opened 2025-08-19
docs: improve gas padding explanation

### LINKED ISSUE #585: Docs: Improve Gas Padding explanation
The current [Gas Padding docs](https://docs.oasis.io/build/sapphire/develop/security#gas-padding) is not explaining in enough depth how it works and how to use it.

The docs should state more explicit that it only protects the code path taken within the contract and that transaction size gets revealed by it. 

Some recommendation when to use it and when not would be helpful and also some guidance to devs that want to masks calldata input size.

uniyalabhishek:
Closes #585 

Preview URL: https://deploy-preview-616--oasisprotocol-sapphire-paratime.netlify.app/build/sapphire/develop/security#gas-padding

### DIFF
--- a/docs/develop/security.md
+++ b/docs/develop/security.md
@@ -1,5 +1,5 @@
 ---
-description: "Recipes for Confidentiality: Security considerations when writing confidential contracts"
+description: Recipes for Confidentiality, Security considerations when writing confidential contracts
 ---
 
 # Security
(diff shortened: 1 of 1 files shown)


docs/develop/security.md
```diff
@@ -21,7 +21,7 @@ npm install -D hardhat-tracer
 and add `hardhat-tracer` to your `config.ts` file,
 
 ```typescript
-import "hardhat-tracer"
+import 'hardhat-tracer';
```
matevz, line 24: I would keep double quotes. We're using those in our examples.

```suggestion
import "hardhat-tracer";
```

docs/develop/security.md
```diff
@@ -1,5 +1,5 @@
 ---
-description: "Recipes for Confidentiality: Security considerations when writing confidential contracts"
+description: 'Recipes for Confidentiality: Security considerations when writing confidential contracts'
```
matevz, line 2: Quotes can be removed here.

```suggestion
description: Recipes for Confidentiality: Security considerations when writing confidential contracts
```

docs/develop/security.md
```diff
@@ -110,32 +110,98 @@ contract Secret {
 
 ## Gas Padding
 
-To prevent leaking information about a particular transaction, Sapphire
-provides a [precompile] for dApp developers to **pad the amount of gas used
-in a transaction**.
+Gas padding lets you equalize **EVM execution** gas across private code paths to reduce side‑channel leakage. Sapphire provides a [precompile] (`Sapphire.padGas`) that burns execution gas so that your function’s execution cost is brought up to a target amount.
```
matevz, line 113: You can add `Sapphire.padGas` link to https://api.docs.oasis.io/sol/sapphire-contracts/contracts/Sapphire.sol/library.Sapphire.html#padgas

docs/develop/security.md
```diff
@@ -110,32 +110,98 @@ contract Secret {
...
+      }
+      ok = true;
+    }
+    // Equalize execution cost across branches. Pads execution only; tx size cost stays visible.
+    Sapphire.padGas(100_000);
+  }
+}
+```
+
+Choose a target that is ≥ the **worst‑case execution** for the function, with a safety margin. You can measure worst‑case cost in tests (e.g., with tracers) and read the current execution gas via [Sapphire.gasUsed()][used].
```
matevz, line 164:
```suggestion
Choose a target that is greater or equal to the **worst‑case execution** for the function, with a safety margin. You can measure worst‑case cost in tests (e.g., with tracers) and read the current execution gas via [Sapphire.gasUsed()][used].
```

docs/develop/security.md
matevz: Maybe add another sentence

```markdown
The gas padding call is usually done somewhere at the end of the executed code to cover all possible execution paths.
```

Because the function doesn't "hardcode" the total gas spent, but just burns it to the specific amount and then continues with the execution and spends gas normally on.

docs/develop/security.md
uniyalabhishek: In the repo config, Prettier is set to enforce single quotes, and it also formats TypeScript code blocks inside Markdown. so it keeps rewriting import to single quotes. I changed the block to text for quick workaround

docs/develop/security.md
uniyalabhishek: the Netlify deploy failed after removing quotes due to Docusaurus: https://app.netlify.com/projects/oasisprotocol-sapphire-paratime/deploys/68bebc2ea855000008cdc6c5, added it back

PR #621, opened 2025-09-09
wagmi-v2: Fix TypeDoc issue

lubej:
- looks like export const due to generic type didn't work

Built upon https://github.com/oasisprotocol/sapphire-paratime/pull/607


lubej:
Tested locally, it does generate `wrapConnectorWithSapphire` function types now.

PR #636, opened 2025-10-09
docs: add danger tag about immutable variables unencrypted in runtime…

njelich:
… bytecode


ptrus:
I was surprised that this had not yet been added to the documentation. I think it's very silly that such critical and straightforward updates are not merged and are blocked on:

>Commits must have verified signatures.

Maybe we should relax the condition a bit, and give developers some time to set up the commit signing before requiring it 

cc @tjanez

PR #664, opened 2025-12-07
wagmi-v2: Fix WalletConnect "Please call connect() before request()" error

lubej:
Recent https://github.com/oasisprotocol/sapphire-paratime/pull/634 introduced some breaking changes for WalletConnect connector. This PR addresses the change, by obtaining the chainId from the provider itself, instead of calling request on the WalletConnect provider. ChainId is already handled internally in WalletConnect provider, so there is no need to call `eth_chainId` method on said provider.

### DIFF
--- a/examples/wagmi-v2/test/wallet-connect.spec.ts
+++ b/examples/wagmi-v2/test/wallet-connect.spec.ts
@@ -1,38 +1,6 @@
 import { expect, test } from "@playwright/test";
-import { createConfig } from "wagmi";
-import {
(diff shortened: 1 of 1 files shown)


pnpm-lock.yaml
```diff
@@ -284,10 +284,10 @@ importers:
     dependencies:
```
lubej: `.lock` file blew up... I would still rather have the types of `@walletconnect/ethereum-provider`, than smaller lock file.

PR #667, opened 2025-12-10
docs/develop/security: Add Simulation attack section

### LINKED ISSUE #654: docs: Write a "Simulation attack" chapter
Write a "Simulation attack" chapter here:
https://docs.oasis.io/build/sapphire/develop/security

The attack goes like this:
1. You have a contract that gives you away private information, if you pay for it. Another example is the encumbered wallet that generates a transaction in a view-call for you to withdraw assets on another chain. Of course, you are allowed to do this, if you actually own the assets.
2. Instead of actually paying for the service/depositing assets through the transaction, you just simulate the call with a 1) transaction sending assets and then 2) requesting the information based on the "simulated" payment you just did.
3. One would think this doesn't work, because you need two transactions to do this and we don't support `eth_callMany`. Well, you can initiate the two calls from a smart contract which can then be called only once.
4. The original sin is that EVM doesn't differentiate between a call or transaction. To prevent simulation attacks, you need to store the block number when the payment was being made. And for the action that is dependent on the payment, you need to require that it happens at least one block after the payment block. The one-block delay is sufficient also for the deposit-withdraw scenario.

Some context:
https://github.com/oasisprotocol/accounting-module/blob/master/solidity/contracts/Accounting.sol#L618

First (vulnerable) version: https://github.com/oasisprotocol/accounting-module/blob/7113b944bfeb2a78f54c82493dee3564e403b458/solidity/contracts/Accounting.sol#L513

matevz:
Fixes #654 

[PREVIEW](https://deploy-preview-667--oasisprotocol-sapphire-paratime.netlify.app/build/sapphire/develop/security#simulation-attack)

A working PoC lives here: https://github.com/oasisprotocol/demo-starter/tree/matevz/simulation-attack/backend

TODO: Move the PoC to a simpler separate demo-simulation-attack repo and add the `:::example` admonition to the docs.

### DIFF
--- a/docs/develop/security.md
+++ b/docs/develop/security.md
@@ -4,10 +4,179 @@ description: Recipes for Confidentiality, Security considerations when writing c
 
 # Security
 
-This page is an ongoing work in progress to support confidential smart contract
-development. At the moment we address safeguarding storage variable access
-patterns and provide best practices for more secure orderings of error checking
-to prevent leaking contract state.
+This page is focused on best practices for writing secure confidential smart 
+contracts. It outlines potential attacks and how we should prevent or mitigate
+them.
+
+## Simulation Attack
+
+Suppose you store a secret in a smart contract which you reveal after some 
+amount of tokens is transferred to the contract:
+
+```solidity title="SecretBox.sol"
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+contract SecretBox {
+  address public immutable owner;
+
+  uint256 public immutable DEPOSIT_AMOUNT;
+  string private secret;
+  address private payer;
+
+  constructor(uint256 amount) {
+    owner = msg.sender;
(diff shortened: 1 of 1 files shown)


docs/develop/security.md
```diff
@@ -4,10 +4,178 @@ description: Recipes for Confidentiality, Security considerations when writing c
...
+
+Running the script above will leak the secret:
+
+```
+Secret: very secret message
+```
+
+### Protection
+
+The Ethereum virtual machine was designed in a way that you cannot detect
```
njelich: Maybe clearer:

"The design of the Ethereum Virtual Machine is such that a smart contract cannot detect whether it is being executed within a finalized transaction vs a static call..."

njelich (changes requested):

njelich (approved):

PR #692, opened 2026-04-10
ci: Run js tests in parallel

matevz:
This PR:
- moves building of clients/js and integrations into a separate job that uploads packages artifacts
- let contracts-test, test-client-js, test-integrations, test-examples and test-examples-playwright re-use these artifacts
- runs all the above tests in parallel. **Reduces the existing contracts-test CI workflow time from 12 to 8 minutes.**
  - 1.5 minute to compile, then the rest of the tests run in parallel.
  - The longest one is contracts-test job that needs 6.5 minutes. The other ones (examples, cilent, integrations) are shorter.
- removes duplicate test:unit from ci-test.yaml, because they're already included in ci-test-js-contracts.yaml.
- bumps examples/encrypted-events sapphire-hardhat to 2.22.3 that reduces poll interval for sapphire-localnet
- bumps nodeJS from 18.x to 24.x in the CI


.github/workflows/ci-playwright.yaml
```diff
@@ -7,8 +7,43 @@ on:
...
+            clients/js/lib
+            integrations/*/dist
+          retention-days: 1
+
+  examples-playwright-js-test:
+    runs-on: ubuntu-latest
+    needs: build-client-integrations
+    strategy:
+      matrix:
+        example: [wagmi-v2, wagmi-v3, encrypted-events]
```
uniyalabhishek: encrypted-events is hardhat-only and already covered by contracts-test via make -C examples, so I'd drop it from the matrix.

uniyalabhishek (changes requested):

uniyalabhishek (approved):

PR #693, opened 2026-04-13
deps: override axios to 1.15.0

lubej:
Should close https://github.com/oasisprotocol/sapphire-paratime/security/dependabot/355

The vulnerability should not affect the Browser, only Node.js runtime.

It should close the following as well https://github.com/oasisprotocol/sapphire-paratime/security/dependabot/354.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
   "pnpm": {
     "overrides": {
       "elliptic": "^6.6.1",
-      "sha.js": "^2.4.12"
+      "sha.js": "^2.4.12",
+      "axios": "^1.15.0"
     }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,6 +7,7 @@ settings:
 overrides:
   elliptic: ^6.6.1
   sha.js: ^2.4.12
+  axios: ^1.15.0
 
 importers:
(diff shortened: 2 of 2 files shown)


lubej:
`axios` is a transitive dependency through `hardhat-gas-reporter`, which is used only in `hardhat-boilerplate` and `onchain-signer`. @matevz can we just remove the dependency from those 2 projects? From first glance I don't see `hardhat-gas-reporter` being used - which would simplify the dependency chain - and effectively the vulnerabilities altogether.

matevz (approved):

lubej:
> `axios` is a transitive dependency through `hardhat-gas-reporter`, which is used only in `hardhat-boilerplate` and `onchain-signer`. @matevz can we just remove the dependency from those 2 projects? From first glance I don't see `hardhat-gas-reporter` being used - which would simplify the dependency chain - and effectively the vulnerabilities altogether.

Seems like removing `hardhat-gas-reporter` isn't as easy, as it is still pulled in by peer dependency from `hardhat-toolbox`. Disabling `auto-install-peers=false` peer installs seems to brake more than it solves.

https://github.com/oasisprotocol/sapphire-paratime/actions/runs/24389980474/job/71233191842

PR #695, opened 2026-04-17
deps: handlebars upgrade

lubej:
Resolves the following:
https://github.com/oasisprotocol/sapphire-paratime/security/dependabot/335

Dependency chain: solidity-coverage → sc-istanbul → handlebars


matevz (approved):

matevz:
Do we need to release new versions of all JS packages too?

lubej:
> Do we need to release new versions of all JS packages too?

I would, let me just clean up all critical alerts.

PR #702, opened 2026-05-07
Add support for cbor2 6.x

matevz:
Fixes #701

### DIFF
--- a/clients/py/sapphirepy/envelope.py
+++ b/clients/py/sapphirepy/envelope.py
@@ -111,7 +111,7 @@ class TransactionCipher:
         return envelope
 
     def _decode_inner(self, plaintext:bytes) -> bytes:
-        inner_result = cast(ResultInner, cbor2.loads(plaintext))
+        inner_result = cast(ResultInner, cbor2.loads(bytes(plaintext)))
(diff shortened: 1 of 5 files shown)


clients/py/sapphirepy/envelope.py
```diff
@@ -111,7 +111,7 @@ def make_envelope(self, plaintext: bytes):
         return envelope
 
     def _decode_inner(self, plaintext:bytes) -> bytes:
-        inner_result = cast(ResultInner, cbor2.loads(plaintext))
+        inner_result = cast(ResultInner, cbor2.loads(bytes(plaintext)))
```
ptrus, line 114: can we add a simple test case for this?

ptrus (approved):

clients/py/sapphirepy/envelope.py
matevz: Decryption is tested at https://github.com/oasisprotocol/sapphire-paratime/blob/01a797d0a0746c1cf88a2de4b0d72be938f99a52/clients/py/sapphirepy/tests/test_deoxysii.py#L95 if that's what you mean?

PR #709, opened 2026-05-20
contracts/contracts: Fix address derivation in subcalls

peternose:


contracts/contracts/Subcall.sol
```diff
@@ -508,7 +506,7 @@ library Subcall {
     function accountsTransfer(address to, uint128 value) internal {
```
matevz: I would rather change the API for consistency and then use it directly in the subcall as above.

```suggestion
    function accountsTransfer(StakingAddress to, uint128 value) internal {
```

contracts/contracts/ConsensusUtils.sol
```diff
@@ -42,51 +36,76 @@ library ConsensusUtils {
...
+        publicAddress = StakingAddressDerivation.fromEd25519(bytes32(pk));
         secretKey = StakingSecretKey.wrap(bytes32(sk));
     }
+}
 
+/**
+ * @title Staking address derivation utilities
+ * @notice Shared Oasis-style staking address derivation helpers.
+ */
+library StakingAddressDerivation {
```
matevz: Any specific reason for introducing a new library? Because this also needs updates through our examples and docs.

peternose: Removed. Now that I have changed the signature of the method `accountsTransfer`, the function is probably useless as I think there is no method available to convert ETH address to staking address. When required, someone can refactor this code and add it.

The test for transfer is still not complete, as I cannot easily verify the balance of a staking account. One option is to add derivation for ETH addresses, or to add query `accounts.Balances`. But that is probably too much for this PR.

matevz (approved):
Let's merge this.

PR #713, opened 2026-06-07
contracts/contracts/opl: Error if endpoint doesn't exist

peternose:


matevz (approved):
I think this check was mandatory as `0x000000` selectors may return weird results. The code is really tricky though. Initially I wanted to fix it to

```
if (epMissing) revert UnknownEndpoint();
```

but then since `seq` was not increased, the client will probably retry indefinitely.

PR #722, opened 2026-07-14
Add Envio to Sapphire indexers

Jordy-Baby:
This adds Envio to the Indexers table on the Sapphire network information page.

Envio is a high-performance indexing framework that turns smart contract events into a queryable GraphQL API. Its HyperIndex supports indexing any EVM chain out of the box, so developers can index Oasis Sapphire (chain ID 23294) using their own RPC as the data source. Indexers can be self-hosted or run on the managed Envio Cloud service.

Website: https://envio.dev/
Docs: https://docs.envio.dev/

### DIFF
--- a/docs/network.mdx
+++ b/docs/network.mdx
@@ -87,6 +87,7 @@ dedicated RPC endpoints, consider the following providers (in alphabetic order):
 
 | Name (Provider)              | Mainnet URL                                            | Testnet URL                                            | Documentation                                                                                             |
 |------------------------------|--------------------------------------------------------|--------------------------------------------------------|-----------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


matevz (approved):
I agree, we can add Envio.

Jordy-Baby:
Hi team, thanks again for merging this back in July.

One thing worth flagging. The entry is still in `docs/network.mdx` on main, but the rendered page at https://docs.oasis.io/build/sapphire/network does not show it. The Indexers table there currently lists Covalent, Goldsky and Oasis Nexus, without the Envio row that this PR added, so the published docs look to be serving an older build of this file.

Is there a separate release or version bump needed for changes here to reach docs.oasis.io? Happy to open anything else needed on the docs side.

Thanks for considering Envio.

PR #724, opened 2026-07-16
sapphire-contracts: Add ROFLable and ROFLableUpgradeable

matevz:
Implements #723 

- `ROFLable` and `ROFLableUpgradeable` similar to OpenZeppelin's `Ownable` and `OwnableUpgradeable`
- `setRoflAppId` is enabled by default and gated by `onlyROFL` (in other words `roflAppId` is mutable).
- If stored `roflAppId` is zero, then `onlyROFL` gate is ignored. Useful for the tests on non-Sapphire chains.

### DIFF
--- a/contracts/contracts/ROFLable.sol
+++ b/contracts/contracts/ROFLable.sol
@@ -1,0 +1,84 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+import {Subcall} from "./Subcall.sol";
+
+/**
+ * @title Base contract for ROFL-authenticated access control
+ * @notice Provides the `onlyROFL` modifier, restricting calls to
+ * transactions signed by an authorized instance of a specific ROFL app.
+ *
+ * See [`ROFLableUpgradeable`](../ROFLableUpgradeable.sol/abstract.ROFLableUpgradeable.md)
+ * for an upgradeable counterpart of this contract.
+ *
+ * #### Example
+ *
+ * ```solidity
+ * contract MyContract is ROFLable {
+ *   constructor(bytes21 inRoflAppId) ROFLable(inRoflAppId) {}
+ *
+ *   function setPrice(uint128 price) external onlyROFL {
+ *     ...
+ *   }
+ * }
+ * ```
+ */
+abstract contract ROFLable {
+    bytes21 private _roflAppId;
+
+    event RoflAppIdUpdated(
+        bytes21 indexed previousRoflAppId,
+        bytes21 indexed newRoflAppId
+    );
+
+    /**
+     * @notice Initializes the contract, authorizing the given ROFL app.
+     * @param initialRoflAppId ROFL app ID.
+     */
+    constructor(bytes21 initialRoflAppId) {
+        _setRoflAppId(initialRoflAppId);
+    }
+
+    /**
+     * @notice Reverts unless the transaction is signed by an authorized
+     * instance of the configured ROFL app.
+     */
+    modifier onlyROFL() {
+        _checkRoflAppId();
+        _;
+    }
+
+    /**
+     * @notice Returns the currently authorized ROFL app ID.
+     */
+    function roflAppId() public view virtual returns (bytes21) {
+        return _roflAppId;
+    }
+
+    /**
+     * @notice Updates the authorized ROFL app ID. Callable only by the
+     * currently authorized ROFL app.
+     */
+    function setRoflAppId(bytes21 newRoflAppId) public virtual onlyROFL {
+        _setRoflAppId(newRoflAppId);
+    }
+
+    /**
+     * @notice Reverts unless the transaction is signed by an authorized
+     * instance of the configured ROFL app.
+     */
+    function _checkRoflAppId() internal view virtual {
+        Subcall.roflEnsureAuthorizedOrigin(_roflAppId);
+    }
+
+    /**
+     * @notice Updates the authorized ROFL app ID. Internal, without access
+     * restriction - gate calls to this with your own access control (e.g.
+     * `onlyOwner`) when exposing it externally.
+     */
+    function _setRoflAppId(bytes21 newRoflAppId) internal virtual {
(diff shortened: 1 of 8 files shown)


contracts/package.json
```diff
@@ -57,6 +57,7 @@
     "@noble/hashes": "1.3.2"
   },
   "dependencies": {
-    "@openzeppelin/contracts": "^5.0.2"
+    "@openzeppelin/contracts": "^5.0.2",
+    "@openzeppelin/contracts-upgradeable": "5.0.2"
```
rube-de:
```
    "@openzeppelin/contracts": "^5.0.2",
```
resolves to 5.3.0 and same time for upgradable is locked to 5.0.2.

I would recommend to have both on same version

contracts/contracts/ROFLableUpgradeable.sol
```diff
@@ -0,0 +1,120 @@
...
+import {Subcall} from "./Subcall.sol";
+
+/**
+ * @title Upgradeable base contract for ROFL-authenticated access control
+ * @notice Upgradeable counterpart of [`ROFLable`](../ROFLable.sol/abstract.ROFLable.md).
+ * Provides the `onlyROFL` modifier, restricting calls to transactions signed by
+ * an authorized instance of a specific ROFL app.
+ *
+ * The zero app ID (`0x0`) disables the check, letting `onlyROFL`-gated
+ * functions be called by anyone. This is intended for tests running on chains
```
rube-de: For testing this is fine, but idk if having that in a production contract is wise.

contracts/contracts/ROFLableUpgradeable.sol
```diff
@@ -0,0 +1,120 @@
...
+    function setRoflAppId(bytes21 newRoflAppId) public virtual onlyROFL {
+        _setRoflAppId(newRoflAppId);
+    }
+
+    /**
+     * @notice Reverts unless the transaction is signed by an authorized
+     * instance of the configured ROFL app.
+     */
+    function _checkRoflAppId() internal view virtual {
+        bytes21 appId = roflAppId();
```
rube-de: is this on purpose not the same as in the ROFLable.sol?
Maybe the modifiers should be in a separate contract to not have divergence.

contracts/contracts/ROFLableUpgradeable.sol
```diff
@@ -0,0 +1,120 @@
...
+        private
+        pure
+        returns (ROFLableStorage storage $)
+    {
+        assembly {
+            $.slot := ROFLableStorageLocation
+        }
+    }
+
+    event RoflAppIdUpdated(bytes21 previousRoflAppId, bytes21 newRoflAppId);
```
rube-de:
```suggestion
    event RoflAppIdUpdated(bytes21 indexed previousRoflAppId, bytes21 indexed newRoflAppId);
```
for better filtering

contracts/package.json
matevz: ~Bumped and pinned to 5.6.1.~

Bumped and pinned to 5.3.0. Later versions depend on `MCOPY` which is not available on `paris`, just `cancun`.

contracts/contracts/ROFLableUpgradeable.sol
matevz: Removed. For testing one can reimplement `_checkRoflAppId()` in the mock now and ignores the check.

contracts/contracts/ROFLableUpgradeable.sol
matevz: Yes, the upgradable version must access app id via getter, because it's stored at a magic location. The non-upgradable can simply use `_roflAppId`.

contracts/contracts/ROFLableUpgradeable.sol
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    event RoflAppIdUpdated(
+        bytes21 indexed previousRoflAppId,
+        bytes21 indexed newRoflAppId
+    );
+
+    /**
+     * @notice Initializes the contract, authorizing the given ROFL app.
+     * @param initialRoflAppId ROFL app ID, or `0x0` to disable the check.
```
rube-de: the natspec is now wrong, no?

contracts/contracts/ROFLableUpgradeable.sol
```diff
@@ -0,0 +1,115 @@
...
+     * @notice Returns the currently authorized ROFL app ID.
+     */
+    function roflAppId() public view virtual returns (bytes21) {
+        ROFLableStorage storage $ = _getROFLableStorage();
+        return $._roflAppId;
+    }
+
+    /**
+     * @notice Updates the authorized ROFL app ID. Callable only by the
+     * currently authorized ROFL app, or by anyone while unset (`0x0`).
```
rube-de: should 0x0 be allowed here?

contracts/contracts/ROFLable.sol
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * @notice Returns the currently authorized ROFL app ID.
+     */
+    function roflAppId() public view virtual returns (bytes21) {
+        return _roflAppId;
+    }
+
+    /**
+     * @notice Updates the authorized ROFL app ID. Callable only by the
+     * currently authorized ROFL app, or by anyone while unset (`0x0`).
```
rube-de: Should 0x0 be allowed here?

contracts/contracts/ROFLable.sol
```diff
@@ -0,0 +1,84 @@
...
+    bytes21 private _roflAppId;
+
+    event RoflAppIdUpdated(
+        bytes21 indexed previousRoflAppId,
+        bytes21 indexed newRoflAppId
+    );
+
+    /**
+     * @notice Initializes the contract, authorizing the given ROFL app.
+     * @param initialRoflAppId ROFL app ID, or `0x0` to disable the check.
```
rube-de: As 0x0 return is removed, the natspec is out of date

rube-de (approved):
Some comment nits

PR #727, opened 2026-08-04
contracts: Add UPUPSUpgradeable

matevz:
Based on https://github.com/oasisprotocol/privana/blob/cee23496e515707fdc832b1df98c08c02eb65b0c/solidity/contracts/lib/UPUPSUpgradeable.sol

Changes to the original contract above:
- remove `acceptProposedUpgrade()` modifier. It's called automatically when `upgradeToAndCall()` is executed.
- clearing `implementationHash` when the proposal is accepted
- `_authorizeProposeUpgrade()` now accepts `address newImplementation` and `uint256 minBlockNumber`
- remove virtual from getters, add `onlyProxy` to `proposeUpgrade`
- tests

Example usage:
```solidity
contract MyContract is UPUPSUpgradeable, OwnableUpgradeable {
  constructor() {
    _disableInitializers();
  }

  function initialize(address _owner) public initializer {
    __Ownable_init(_owner);
    __UPUPSUpgradeable_init();
  }

  // Gate _authorizeProposeUpgrade with appropriate modifier.
  function _authorizeProposeUpgrade(address, uint256) internal override onlyOwner {}

  // Gate UUPSUpgradeable._authorizeUpgrade with appropriate modifier as you normally do.
  function _authorizeUpgrade(address) internal override onlyOwner {}
}
```

Then in the upgrade script, using `hardhat-upgrades` instead of calling the built-in `upgradeProxy()` helper you need to split it to `prepareUpgrade()`, obtain the implementation address, and then `proposeUpgrade()` and `upgradeToAndCall()`:

```ts
    // Deploy new implementation.
    const deployTx = await hre.upgrades.prepareUpgrade(myContractAddr, MyContract, {
      kind: 'uups',
      constructorArgs: [],
      redeployImplementation: 'always',
      txOverrides: { gasLimit: 15000000 },
      getTxResponse: true,
    }) as TransactionResponse;
    const deployReceipt = await deployTx.wait();
    const newImplAddress = deployReceipt!.contractAddress!;

    // Propose.
    const txProposeUpgrade = await (await current.proposeUpgrade(newImplAddress, 0)).wait();
    
    // Upgrade.
    const txUpgradeAndCall = await (await current.upgradeToAndCall(newImplAddress, "0x")).wait();

### CODE REVIEW

File: contracts/contracts/UPUPSUpgradeable.sol [hunk 1]
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * @notice Initializes the contract.
+     */
+    function __UPUPSUpgradeable_init() internal onlyInitializing {}
+
+    /**
+     * @notice Checks if the proposed upgrade is valid and clears the proposed
+     * upgrade bits.
+     */
+    modifier acceptProposedUpgrade(address newImplementation) {
```
uniyalabhishek: proposal enforcement is optional here. a subclass can use the standard OZ `onlyOwner` upgrade hook and upgrade without any proposal. I reproduced this locally.

let's enforce proposal acceptance in a non-overridable base path, so omitting this modifier cannot disable the protection.

uniyalabhishek (approved):

contracts/contracts/UPUPSUpgradeable.sol
matevz: Good idea. I put the `acceptProposedUpgrade()` inside the `upgradeAndCall()` now so there is no need for the modifier anymore.

matevz: Might be a good idea to rethink again security concerns, whether someone can silently override `upgradeToAndCall()` and thus ignore the `acceptProposedUpgrade()` check.

contracts/contracts/UPUPSUpgradeable.sol
```diff
@@ -0,0 +1,199 @@
...
+        if (minBlockNumber <= block.number) {
+            revert MinBlockNumberInPast();
+        }
+
+        UPUPSUpgradeableStorage storage $ = _getUPUPSUpgradeableStorage();
+        $._newImplementation = newImplementation;
+        $._newImplementationHash = newImplementation.codehash;
+        $._minBlockNumber = minBlockNumber;
+
+        _authorizeProposeUpgrade();
```
rube-de: this violates a bit CEI. The authorize check should be done before writing to storage.

```suggestion
        _authorizeProposeUpgrade();(newImplementation, minBlockNumber);

        UPUPSUpgradeableStorage storage $ = _getUPUPSUpgradeableStorage();
        $._newImplementation = newImplementation;
        $._newImplementationHash = newImplementation.codehash;
        $._minBlockNumber = minBlockNumber;
```

OZ 5.4.0’s `upgradeToAndCall` also calls  `_authorizeUpgrade(newImplementation)` first before touching storage.

contracts/contracts/UPUPSUpgradeable.sol
```diff
@@ -0,0 +1,199 @@
...
+ *     __Ownable_init(_owner);
+ *   }
+ *
+ *   // Gate proposeUpgrade with appropriate modifier.
+ *   function _authorizeProposeUpgrade() internal onlyOwner { }
+ *
+ *   // Gate UUPSUpgradeable.authorizeUpgrade with additional acceptProposedUpgrade() modifier.
+ *   function _authorizeUpgrade(address newImpl) internal onlyOwner acceptProposedUpgrade(newImpl) { }
+ * }
+ * ```
```
rube-de, line 32: This example fails to compile as it misses override on the _authorize functions.
Also should maybe add `_disableInitializers()` and `__UPUPSUpgradeable_init()`

